PR #8, opened 2016-11-10
Fix convert command versioning.

### LINKED ISSUE #7: zapier convert command creates outdated package.json file
In particular, the version of `zapier-platform-core` is pegged to 0.3.10, which predates a lot of goodies like `zapier.tools.*`.

bryanhelmig:
Should fix #7 as reported by @barillax.

@stevemolitor heads up!


bryanhelmig:
I'm not sure why `should print a nice little table` is failing so hard... :-/

Works fine locally.

PR #27, opened 2017-02-24
Use 'archiver' to create the zip file

benzapier:
After running into some strange edge-cases where AdmZip was creating a corrupt Zip file, the `archiver` npm seems to produce more consistent results.


bryanhelmig:
Very nice - were you able to confirm the resulting zip works fine uploaded @benzapier?

benzapier:
Yup, the zip file created using `archiver` worked much better!

package.json
```diff
@@ -28,6 +28,7 @@
   },
   "dependencies": {
     "adm-zip": "0.4.7",
+    "archiver": "^1.3.0",
```
BrunoBernardino: Can we please make the version static i.e. `"1.3.0"` instead of `"^1.3.0"`?

PR #29, opened 2017-02-25
Format markdown for console

FokkeZB:
Saturday morning boredom... formats the docs Markdown for console:

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "colors": "1.1.2",
     "fs-extra": "0.30.0",
     "lodash": "4.15.0",
+    "msee": "^0.3.3",
     "node-fetch": "1.6.1",
     "read": "1.0.7",
     "string-length": "1.0.1",
--- a/src/utils/display.js
+++ b/src/utils/display.js
@@ -4,18 +4,15 @@ const readline = require('readline');
 // to simplify the columns/tables? the | - decoration is big
 const Table = require('cli-table2');
 const colors = require('colors/safe');
+const msee = require('msee');
 const stringLength = require('string-length');
 const _ = require('lodash');
 const read = require('read');
 
 const notUndef = (s) => String(s === undefined ? '' : s).trim();
 
-const unBacktick = (s) => s.replace(/\n?`+(bash)?/g, '');
-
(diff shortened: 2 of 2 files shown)


bryanhelmig:
My only concern is the size of the now required markdown library. However, that is minimized considering it is isolated to the CLI and does not leak into core (which would explode the build).

```
$ weigh msee
Approximate weight of msee:
  Uncompressed: 370 kB
  Minified (uglify): 165 kB
  Minified and gzipped (level: default): 52.5 kB
```

FokkeZB:
There's also [marked-terminal](https://www.npmjs.com/package/marked-terminal), with more options and ~ weight:

```
Approximate weight of marked-terminal:
  Uncompressed: 378 kB
  Minified (uglify): 188 kB
  Minified and gzipped (level: default): 50.6 kB
```

xavdid:
I was thinking about this today and would argue we don't need it. Per the [original spec](https://daringfireball.net/projects/markdown/): 

> The idea is that a Markdown-formatted document should be publishable as-is, as plain text

The presence of the markdown should be self sufficient without needing extra formatting. Text like `**required**` stands out because of the asterisks without needing to actually be bold. 

Also this is totally a personal preference thing so I'll understand if we include formatting anyway. Just want to add my 2¢.

FokkeZB:
@xavdid Surely `***required**` stands out, but you don't see Markdown used in CLI outputs often and if you don't know it's markdown (unlike when you view a MD document, when you know it from the context - filename/viewer) it can be confusing:

BrunoBernardino:
I like the colors, but personally think it can cause more problems (it messes with color blindness, accessibility, etc.).

If this can have an opt-out option (like `--plain` for the non-colorized version) I'm all for it.

FokkeZB:
@BrunoBernardino excellent idea to turn `.zapierrc` into a config file rather than just storing the `deployKey`. That would probably take some reorganization though, since it's tight into `utils/api.js` atm.

FokkeZB:
Closing stale PR. We can circle back once we have bandwidth and maybe tie it into introducing a `~/.zapierrc` config file.

PR #36, opened 2017-03-08
Check node version & init w/.nvm rc

### LINKED ISSUE #34: Zapier Validate should verify the Node version used matches production
I ran `zapier validate` and `push` with success, but then when I used it got:

xavdid:
Fixes #34

### DIFF
--- a/README-source.md
+++ b/README-source.md
@@ -14,16 +14,11 @@ Zapier is a platform for creating integrations and workflows. This CLI is your g
 
 ## Requirements
 
-The Zapier CLI and Platform require Node.js `v4.3.2`. We recommend using [nvm](https://github.com/creationix/nvm) to manage your Node.js installation.
+All Zapier CLI apps are run using Node.js `LAMBDA_VERSION`.
 
-On Mac (via [homebrew](http://brew.sh/)):
+You can develop using any version of Node you'd like, but your code has to run on Node `LAMBDA_VERSION`. Most developers will accomplish this by developing on their preferred version and then transpiling their with [Babel](https://babeljs.io/) (or similar).
 
-```bash
-brew install nvm
-nvm install v4.3.2
-nvm use v4.3.2
-```
-
+To ensure stability for our users, we also require that you run your tests on `LAMBDA_VERSION` as well. If you don't have it available, we recommend using either [nvm](https://github.com/creationix/nvm#installation) or [n](https://github.com/tj/n#installation) to install `LAMBDA_VERSION` and run the tests locally. In the case of NVM, you can use `nvm exec LAMBDA_VERSION zapier test` so you can run passing tests without having to switch versions while developing.
 
 ## Tutorial
 
--- a/src/commands/test.js
+++ b/src/commands/test.js
@@ -1,5 +1,8 @@
+'use strict';
+
 const _ = require('lodash');
 const utils = require('../utils');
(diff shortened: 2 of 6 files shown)


xavdid:
`npm test` takes a line of shell code, so we can set them up with any defaults we want (and have stuff in the example repos) but this is the only way to enforce it at a code level. doesn't work if they run `npm test` vs `zapier test`, but I don't think there's a way to integrate with whatever test handler they're running. 

We could ship with mocha and in the `test` command run `mocha ourfile.js`, but that seems like a much heavier solution.

The other option is to not enforce it, but give them a default `.travis.yml` file, which'll test it on 4.3.2. That's pretty out of the way though, and wouldn't have caught Fokke's initial issue (that we're solving for)

src/commands/test.js
```diff
@@ -11,6 +14,10 @@ const test = (context) => {
     extraEnv.DETAILED_LOG_TO_STDOUT = 'true';
   }
 
+  if (process.version !== LAMBDA_VERSION) {
+    throw new Error(`You're running tests on Node ${process.version}, but Zapier runs your code on ${LAMBDA_VERSION}. To ensure errors don't sneak up on you, the version numbers must match.`);
```
FokkeZB: Since we include a `.nvmrc` I'd add something like "Install and run `nvm use` to switch (https://github.com/creationix/nvm)."

xavdid: I don't love pushing NVM as a requirement since it's not (just a suggestion to manage versions). I'd rather have something in the readme that goes into the version requirements and give NVM as a suggestion (along with [n](https://github.com/tj/n) and any other good tools). 

The error message should specify what the error is (version mismatch) and then could point to the docs for resolution suggestions). 

How about `You're running tests on Node ${process.version}, but Zapier runs your code on ${LAMBDA_VERSION}. Your test version must match. See ${README_LINK} for suggestions.`

FokkeZB: Sounds reasonable. Lets add that README section then :)

xavdid: updated!

xavdid:
Changed the language in the `Requirements` section of the readme a little. @bryanhelmig you cool with that language? NVM apparently doesn't officially support `brew` installation (though I think that's how a lot of people use it) so I figured it's better to link to their installation docs than specify it ourselves.

README-source.md
```diff
@@ -14,16 +14,11 @@ Zapier is a platform for creating integrations and workflows. This CLI is your g
...
-On Mac (via [homebrew](http://brew.sh/)):
-
-```bash
-brew install nvm
-nvm install v4.3.2
-nvm use v4.3.2
-```
+You can develop using any version of Node you'd like, but your code has to run on Node `LAMBDA_VERSION`. Most developers will accomplish this by developing on their preferred version and then transpiling their with [Babel](https://babeljs.io/) (or similar).
 
+To ensure stability for our users, we also require that you run your tests on `LAMBDA_VERSION` as well. If you don't have it available, we recommend using either [nvm](https://github.com/creationix/nvm#installation) or [n](https://github.com/tj/n#installation) to install `LAMBDA_VERSION` and run the tests locally. In the case of NVM, you can use `nvm exec LAMBDA_VERSION zapier test` so you can run passing tests without having to switch versions while developing.
```
FokkeZB: Didn't know this, awesome!

PR #51, opened 2017-03-27
Use BASE_ENDPOINT instead of hard coding zapier.com in validation text

### LINKED ISSUE #21: Rely on ZAPIER_BASE_ENDPOINT when printing output
Right now commands like `zapier convert` and `zapier push` are hard-coded to print messages that refer to `zapier.com`. Would be helpful for local development and also beta testing if it correctly checked ZAPIER_BASE_ENDPOINT.

alexleventer:
Fixes #21 in the CLI - I did not make this change in the docs.

### DIFF
--- a/src/commands/convert.js
+++ b/src/commands/convert.js
@@ -11,7 +11,7 @@ const convert = (context, appid, location) => {
 
   appid = Number(appid);
   if (!appid) {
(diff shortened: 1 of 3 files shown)


BrunoBernardino:
Thank you so much for the PR @alexleventer ! I'm fixing #21 as part of #56 as well, so this won't be merged.

I've sent you a token of appreciation to your email!

PR #55, opened 2017-03-28
Do not show command help on error

### LINKED ISSUE #30: Don't print help docs on error.
It obscures the error message:

```bash
$ zapier init
Initializes a new Zapier app in a directory.

Usage: `zapier init path`

Initializes a new Zapier app. If you specify a template, will download and install the app from that template.

After running this, you'll have a new example app in your directory. If you re-run this command on an existing directory it will leave existing files alone and not clobber them.

> Note: this doesn't register or deploy the app with Zapier - try zapier register "Example" and zapier push for that!

**Arguments**

* location [value] -- **required**,
* --template={minimal,resource,trigger,create,search,middleware,basic-auth,custom-auth,oauth2,session-auth,babel} -- _optional_, select a starting app template. Default is minimal

$ zapier init example-app --template=minimal
# Let's initialize your app!
#
#   Downloading zapier/zapier-platform-example-app-minimal starter app - done!
#   Copy /users/username/code/example-app/.gitignore - done!
#   Copy /users/username/code/example-app/index.js - done!
#   Copy /users/username/code/example-app/package.json - done!
#   Copy /users/username/code/example-app/test/index.js - done!
#
# Finished! You might need to npm install then try zapier test!

Errors running command `zapier init`:

!!!   Missing required positional argument 1/location
```

Probably better would be:

```
$ zapier init
Errors running command `zapier init`. Try `zapier help init` for more information.

!!!   Missing required positional argument 1/location
```

BrunoBernardino:
Show help suggestion instead, and for `init` use "path" instead of "location" as the argument name.

Fixes #30.

Shows like this instead:

screen shot 2017-03-28 at 14 28 57


src/entry.js
```diff
@@ -71,20 +71,20 @@ module.exports = (argv) => {
...
+    context.line();
+    errors.forEach((error) => context.line(colors.red(`!!!   ${error}`)));
+    context.line();
+    context.line(`For more information, run \`zapier help ${command}\`.`);
+    context.line();
+    process.exit(1);
   }
 
   commandFunc.apply(commands, [context].concat(args))
     .then(() => {
```
ibolmo: I wonder if this is still necessary after the removal.

BrunoBernardino: Great question! Yeah, otherwise the spinner would continue when the command finishes successfully.

BrunoBernardino:
@FokkeZB yeah. I think `path` is more intuitive in the context here.

PR #57, opened 2017-03-30
Parse outputFields for `zapier convert`

### LINKED ISSUE #23: zapier convert is not converting Sample Data correctly
`zapier convert XXX` is producing Sample Data like so:

```
module.exports = {
  key: 'new_completed_ride',
  noun: 'Completed Ride',

  display: {
    label: 'New Completed Ride',
    description: 'Triggers when you complete a ride.'
  },

  operation: {
    inputFields: [

    ],
    sample: {
      destination__address: {
        type: 'string',
        label: 'undefined'
      },
      destination__eta_seconds: {
        type: 'string',
        label: 'undefined'
      },
      destination__lat: {
        type: 'string',
        label: 'undefined'
      },
      destination__lng: {
        type: 'string',
        label: 'undefined'
      },
      driver__first_name: {
        type: 'string',
        label: 'undefined'
      }
    },

    perform: triggerNewcompletedride
  }
};
```

This should not be the case but instead something along the lines of `{vehicle__model: 'example text'}`

BrunoBernardino:
Fixes #23

To really create a `sample`, we'd need to get info in the v2 definition dump that we currently don't have.

### DIFF
--- a/src/utils/convert.js
+++ b/src/utils/convert.js
@@ -102,7 +102,7 @@ const renderSampleField = (def) => {
 const renderSample = (definition) => {
   const fields = _.map(definition.sample_result_fields, renderSampleField);
 
-  if (!fields || fields.length === 0) {
+  if (!fields.length) {
     return '';
   }
 
--- a/src/utils/convert.js
+++ b/src/utils/convert.js
@@ -83,20 +83,32 @@ ${props.join(',\n')}
 };
 
 const renderSampleField = (def) => {
-  const type = typesMap[def.type];
+  const type = typesMap[def.type] || 'string';
 
-  return `      ${def.key}: {
+  if (def.label) {
+    return `      {
+        key: ${quote(def.key)},
         type: ${quote(type)},
         label: ${quote(def.label)}
+      }`;
+  }
+
+  return `      {
+        key: ${quote(def.key)},
+        type: ${quote(type)}
       }`;
 };
 
 const renderSample = (definition) => {
   const fields = _.map(definition.sample_result_fields, renderSampleField);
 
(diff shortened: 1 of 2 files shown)


src/utils/convert.js
```diff
@@ -83,20 +83,32 @@ ${props.join(',\n')}
 };
 
 const renderSampleField = (def) => {
-  const type = typesMap[def.type];
+  const type = typesMap[def.type] || 'string';
 
-  return `      ${def.key}: {
+  if (def.label) {
```
ibolmo: Perhaps:

```js
return JSON.dump(_.mapValues(_.merge({}, def, {type}), (value) => quote(value)), null, 2);
```

Note that should they have extra keys than we needed they'll be included. I wonder if we should permit it.

src/utils/convert.js
```diff
@@ -83,20 +83,32 @@ ${props.join(',\n')}
...
+        key: ${quote(def.key)},
+        type: ${quote(type)}
+      }`;
 };
 
 const renderSample = (definition) => {
   const fields = _.map(definition.sample_result_fields, renderSampleField);
 
-  return `    sample: {
+  if (!fields || fields.length === 0) {
```
ibolmo: Looks like lodash will always return an array. 

```js
if (!fields.length) {
  return '';
}
```

Any particular reason we're not using `JSON.dump`?

src/utils/convert.js
xavdid: JSON.dump seems like a good approach here, but I'd be careful with the big 1 liner - with all the parens and commas I find that very hard to grok on a skim.

PR #58, opened 2017-03-30
Improves validate-templates and zapier test commands

BrunoBernardino:
- `zapier test` now supports a `--timeout=X` for X milliseconds added to mocha.
- `npm run validate-templates` now adds a timeout for mocha, and builds the babel example before testing.

### DIFF
--- a/scripts/validate-app-templates.js
+++ b/scripts/validate-app-templates.js
@@ -13,7 +13,7 @@ const appTemplates = require('../lib/app-templates');
 const validateAppTemplate = (template, rootTmpDir) => {
   //const appDir = path.resolve(rootTmpDir, template);
   const zapierCmd = path.resolve(__dirname, '../zapier.js');
-  const extraCmd = (template === 'babel') ? ' && rm -rf lib && node_modules/babel-cli/bin/babel.js src --out-dir lib' : '';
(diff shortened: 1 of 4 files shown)


scripts/validate-app-templates.js
```diff
@@ -13,6 +13,7 @@ const appTemplates = require('../lib/app-templates');
 const validateAppTemplate = (template, rootTmpDir) => {
   //const appDir = path.resolve(rootTmpDir, template);
   const zapierCmd = path.resolve(__dirname, '../zapier.js');
+  const extraCmd = (template === 'babel') ? ' && rm -rf lib && node_modules/babel-cli/bin/babel.js src --out-dir lib' : '';
```
ibolmo: `rm -rf` is a bit scary

xavdid: +1. I know we made that folder and it should be totally fine, but users do [weird things](https://xkcd.com/1172/). I think we can safely re-run babel into an existing directory and it'll update properly.

BrunoBernardino: I was waiting on https://github.com/zapier/zapier-platform-example-app-babel/pull/2 to get in. Now I can change that to simply `npm run zapier-build`.

PR #62, opened 2017-04-07
Updating changelog

### LINKED ISSUE #61: Update the changelog
The `CHANGELOG` seems to be out of date.

Can you keep this in sync with any changes to make it easier to review for potential issues when updating?

BrunoBernardino:
Fixes #61

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,42 @@
+## 1.0.7
+
+* Improved `link` command UX on error/non-existing apps.
+
+## 1.0.6
+
+* Fixed `convert` command for `outputFields`.
+* Improvements on `validate-templates` and `zapier test` (now supports `--timeout=5000`).
+* Minor docs fixes.
+
+## 1.0.5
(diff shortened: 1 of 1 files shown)


xavdid:
going forward, do we want to do something like [this](https://coderwall.com/p/5cv5lg/generate-your-changelogs-with-git-log) to update the changelog as part of the `version` command?

BrunoBernardino:
Great idea! Personally, I like giving the changelog a more human/manual touch, though something like this can happen (forgetting to update it). I'm open to implementing something like that.

bcooksey:
Cool idea. If we combined that with commit squashing I think that would work out well. 

Would there be value in tweaking that command so we can also have the ability to manually write a message when need be?

xavdid:
I believe you can edit the file manually after generation and it won't be
overwritten by future releases (though that probably depends on the tool
used)

~DavidB
[davidbrownman.com]
c: 303.819.0850

PR #65, opened 2017-04-12
Prevent stacktrace from showing for test command.

kola-er:
Instead of raising an error for bad node version, print the error message only.

### DIFF
--- a/src/commands/test.js
+++ b/src/commands/test.js
@@ -16,7 +16,9 @@ const test = (context) => {
   }
 
   if (process.version !== LAMBDA_VERSION) {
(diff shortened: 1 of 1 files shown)


src/commands/test.js
```diff
@@ -16,7 +16,9 @@ const test = (context) => {
   }
 
   if (process.version !== LAMBDA_VERSION) {
-    throw new Error(`You're running tests on Node ${process.version}, but Zapier runs your code on ${LAMBDA_VERSION}. The version numbers must match. See https://github.com/zapier/zapier-platform-cli#requirements for more info.`);
+    context.line(`You're running tests on Node ${process.version}, but Zapier runs your code on ${LAMBDA_VERSION}. The version numbers must match. See https://github.com/zapier/zapier-platform-cli#requirements for more info.`);
+    process.exitCode = 1;
+    return new Promise(() => {});
```
ibolmo: I believe you could do `Promise.resolve()` which should be equivalent.

PR #67, opened 2017-04-13
Fixes middleware snippet

ibolmo:


BrunoBernardino:
The doc fix needs to be in `README-source.md`, but don't worry, I'll get to it later.

ibolmo:
Actually there's no mention of `afterRequest` in `README-source.md`. The `README-source.md`, though, does a file inclusion/import of this snippet.

PR #73, opened 2017-04-20
Add support for node v6.10.2

kola-er:
Documents support for node v6.10.2 in reference to [PR](https://github.com/zapier/zapier/pull/11631).


README-source.md
```diff
@@ -67,7 +67,7 @@ brew install nvm
 nvm install v4.3.2
```
BrunoBernardino: Might want to update this to show v6 as a default, I suppose?

docs/cli.html
```diff
@@ -693,7 +693,7 @@ <h2 id="init">init</h2>
 <p>Note: this doesn&apos;t register or deploy the app with Zapier - try <code>zapier register &quot;Example&quot;</code> and <code>zapier push</code> for that!</p>
 </blockquote><p><strong>Arguments</strong></p><ul>
 <li><code>path [value]</code> -- <strong>required</strong>,</li>
-<li><code>--template={minimal,resource,trigger,create,search,middleware,basic-auth,custom-auth,oauth2,session-auth,babel,rest-hooks,files}</code> -- <em>optional</em>, select a starting app template. Default is <code>minimal</code></li>
+<li><code>--template={minimal,resource,trigger,create,search,middleware,basic-auth,custom-auth,oauth2,session-auth,babel,rest-hooks}</code> -- <em>optional</em>, select a starting app template. Default is <code>minimal</code></li>
```
BrunoBernardino: Uh oh, something's wrong here. Did you have the latest `master`? This should include `files`.

kola-er: I'm not sure why `files` was removed, but I'm sure it's rebased with `master`. I couldn't revert it back by running the command that generated it: `npm run docs`. I just manually added it back. What do you think @BrunoBernardino?

README-source.md
kola-er: Done

docs/cli.html
BrunoBernardino: Something's funky, then. I might've messed something up?

docs/cli.html
```diff
@@ -1147,7 +1147,8 @@ <h2 id="test">test</h2>
       <blockquote>
 <p>Tests your app via <code>npm test</code>.</p>
 </blockquote><p>  <strong>Usage:</strong> <code>zapier test</code></p><p>  This command is effectively the same as <code>npm test</code>, except we wire in some custom tests to validate your app. We recommend using mocha as your testing framework.</p><p><strong>Arguments</strong></p><ul>
-<li><code>--debug</code> -- <em>optional</em>, print zapier detailed logs to standard out</li>
+<li><code>--quiet</code> -- <em>optional</em>, do not print zapier detailed logs to standard out</li>
```
BrunoBernardino: This is strange. It's like the changes from https://github.com/zapier/zapier-platform-cli/pull/71 aren't included in this PR.

README-source.md
```diff
@@ -64,10 +64,10 @@ For NVM on Mac (via [homebrew](http://brew.sh/)):
 
 ```bash
 brew install nvm
-nvm install v4.3.2
+nvm install v6.10.2
 ```
 
-You can then either swap to that version with `nvm use v4.3.2`, or do `nvm exec LAMBDA_VERSION zapier test` so you can run tests without having to switch versions while developing.
```
xavdid: You shouldn't remove `LAMBDA_VERSION` here, IIRC that's how we do the version replace

src/commands/test.js
```diff
@@ -13,7 +13,7 @@ const test = (context) => {
     extraEnv.DETAILED_LOG_TO_STDOUT = 'true';
   }
 
-  if (process.version !== LAMBDA_VERSION) {
+  if ( !LAMBDA_VERSION.includes(process.version)) {
```
xavdid: This will also need to revert when update since we'll be back to a single version

kola-er:
@xavdid Yes, currently users' code runs on only `v4.3.2`. But we want to allow users' code to be able to run on `v6.10.2` as well. That's the reason why `LAMBDA_VERSION` is a collection of the two. With this, either of the version can be chosen.

Or are you suggesting that we support just one version? For instance, make `LAMBDA_VERSION = 'v6.10.2'`.

xavdid:
Correct, since we only run the code in one place on one version, only one should be supported. We have to wait for the server change before we can save it here though

kola-er:
@xavdid True, we only run the code in one place on one version currently. With this [PR](https://github.com/zapier/zapier/pull/11631) that I referenced here in the description, the code can be run on either of `v4.3.2` or `v6.10.2` in one place at any point in time. [This](https://github.com/zapier/zapier/blob/bad237cb40ed0db641f55b7963b9e2a12ca38562/lambda_pools/utils.py#L219-L227) and [that](https://github.com/zapier/zapier/blob/bad237cb40ed0db641f55b7963b9e2a12ca38562/lambda_pools/models.py#L168-L169) explain it.

Please if you still feel we should support just one version, add your review to [PR](https://github.com/zapier/zapier/pull/11631).

Yes, I agree that [PR](https://github.com/zapier/zapier/pull/11631) has to be merged before this one. I did it alongside not to forget.

kola-er:
Okay @xavdid, we could make `LAMBDA_VERSION` just `v6.10.2` to make new cli apps use v6.10.2 only. But the backend would still support v4.3.2 for those old apps that refuse to upgrade. Please, are we in agreement on this?

xavdid:
I think that is reasonable? I don't know as much about how cli apps run, but if y'all are on top if it's probably fine. In my experience, that variable is literally the one point of communication we have between the lambda env and the developer.

I've got a few questions:

1. Is there a way for a developer to choose which lambda version their code runs on? 
2. When we release node 6, is there a way to migrate cli code that's currently on 4?

We should set `LAMBDA_VERSION` to a single version that is the default for new projects. If there's a way to check how your code runs, we could also put a blurb about how old code is currently running on node 4 and `here` is how you update that, should you choose to.

kola-er:
@xavdid If my understanding is correct, we use `LAMBDA_VERSION` variable to inform developers the Lambda NodeJS version their code will run on when end users' Zaps are ON - no communication with AWS Lambda during development.

Answers to the two questions:
1. The backend depends on each CLI app's package.json file to determine the lambda version to use. So, yes developers can choose which lambda version to run their code on by setting the NodeJS version in their app's package.json file to either 4.3.x or 6.10.x. Anything different from the two versions would prevent their app from being uploaded at the run of `zapier push`.

2. Yes, developers would be able to migrate their CLI apps running on Lambda NodeJS version 4.3 to run on Lambda NodeJS version 6.10 by simply updating their apps' NodeJS version in the package.json file from 4.3.x to 6.10.x. The backend would do the actual migration at the next run of `zapier push`.
cc @BrunoBernardino @bcooksey @bryanhelmig Please, help me validate my understanding.

I don't know if there's a way to check how a CLI app code runs on Lambda. I think you wanted to give me a link to how to update a blurb, but it isn't a link. Please, I would love to see it.

BrunoBernardino:
I think we could determine CLI `>=1.1.x` runs on `v6.10.2`, and `<=1.0.x` runs on `v4.3.2` and have that be the decider, and when we launch this, do a `npm version minor`.

If they upgrade the core version, they need to push a new version (unless it's not global), and for that the checks and validations would run the new code in v6 so I believe we should be safe.

I also prefer that over the complexity of allowing two simultaneous runtime versions.

I don't see why someone would prefer to run stuff on `v4` while having `v6` available.

kola-er:
Cool. I like this approach too.

For the upgrade of the core version for the existing global app, I think it could still be achieved without pushing a new version of the app. Running `zapier push` over a global app validates before uploading, right? And part of the validation is to run the code in Lambda on the NodeJS version specified, right? @BrunoBernardino

BrunoBernardino:
I'm not sure I understand what you mean @kola-er . An app version in `production` (global) can't be changed at all (`push` won't work). A new version needs to be created, and if it's not `production` (i.e. promoted), then it can be updated, yeah.

kola-er:
Got it now. Thanks @BrunoBernardino

bcooksey:
Thinking beyond this upgrade, do we want to adopt a pattern for CLI where node version X only works with CLI Y to Z? Node 8 becomes the active LTS in August, so depending on how quickly Lambda adopts support for it, we could be looking at a very narrow range of versions for Node 6, then we'd bump to 8.

It would make the CLI easier to maintain. It has a downside though that if a CLI app uses patterns or libraries that are not compatible with the next Node version, we lock that app out of all new CLI features until the app upgrades, which may include waiting around for libraries to upgrade.

xavdid:
node 6 hit LTS last Oct and they just recently got support for it. I assume there'll be similar lead times for 8?

BrunoBernardino:
You bring up a great point @bcooksey though I'd prefer making it more predictable and easier to maintain. We always allow support for "newer" stuff through `babel`.

kola-er:
Just waiting on whatever we decide.

bcooksey:
If that is the way everybody is leaning, let's make it a major version bump, as I don't think we can update the major node version a dev's code is running on and call that backward compatible. So:

CLI 1.X - node 4.3
CLI 2.X - node 6.10

If we later decide we want to make core support multiple node runtimes, those could always be minor bumps. Then if we ever want to drop a node runtime, we just do another major version bump.

Sound reasonable?

kola-er:
jekins retest frontend

PR #87, opened 2017-06-07
Update Resource docs about key autogeneration

BrunoBernardino:
Tagging @ibolmo since you noticed this was missing.

### DIFF
--- a/README-source.md
+++ b/README-source.md
@@ -410,6 +410,8 @@ Adding a create method looks very similar.
 Every method you define on a `resource` Zapier converts to the appropriate Trigger, Create, or Search. Our examples
 above would result in an app with a New Recipe Trigger and an Add Recipe Create.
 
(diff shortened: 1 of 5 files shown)


README-source.md
```diff
@@ -410,6 +410,8 @@ Adding a create method looks very similar.
 Every method you define on a `resource` Zapier converts to the appropriate Trigger, Create, or Search. Our examples
```
ibolmo: Perhaps move the "Note" from below up to here.. maybe a sidebar:

# Note
>> Every method you define on a `resource` Zapier converts to the appropriate Trigger, Create, or Search. Our examples.

>> Moreover, Trigger, Create, Search, and Search or Create are automatically generated (in case you want to use them in a dynamic dropdown), like: `{resourceName}List`, `{resourceName}Create`, `{resourceName}Search`, and `{resourceName}SearchOrCreate`; in the examples above, `{resourceName}` would be `recipe`.

BrunoBernardino: Thanks for the suggestion @ibolmo ! I've kept it as is, as I don't think it's _that_ important to showcase this, over other nuances. We'll see if people miss it, we'll make it more obvious.

ibolmo: @BrunoBernardino how will you check if people are missing this?

BrunoBernardino: I'm thinking we'll note if people start asking about this.

PR #88, opened 2017-06-07
Version checks

xavdid:
Per discussion [here](https://zapier.slack.com/archives/C03TB1UMA/p1496344685110656), check version in 2 ways:

* during validation and testing. plus that function is broken out into a function, so we can add it wherever
* add the super cool [update-notifier](https://github.com/yeoman/update-notifier), which bugs the user daily (configurable, that's the default) if there's a newer version on npm that they're not using.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,8 @@
     "read": "1.0.7",
     "string-length": "1.0.1",
     "through2": "2.0.1",
-    "tmp": "0.0.28"
+    "tmp": "0.0.28",
+    "update-notifier": "^2.2.0"
   },
   "devDependencies": {
     "babel-cli": "6.14.0",
--- a/src/commands/test.js
+++ b/src/commands/test.js
@@ -13,7 +13,7 @@ const test = (context) => {
     extraEnv.DETAILED_LOG_TO_STDOUT = 'true';
   }
 
-  if (!utils.correctVersion()) {
+  if (!utils.isCorrectVersion(context)) {
     process.exitCode = 1;
     return Promise.resolve();
   }
--- a/src/commands/test.js
+++ b/src/commands/test.js
@@ -3,7 +3,7 @@
 const _ = require('lodash');
 const constants = require('../constants');
 const utils = require('../utils');
-const LAMBDA_VERSION = 'v4.3.2';
+
 
 const test = (context) => {
   const extraEnv = {};
@@ -13,8 +13,7 @@ const test = (context) => {
     extraEnv.DETAILED_LOG_TO_STDOUT = 'true';
   }
 
-  if (process.version !== LAMBDA_VERSION) {
-    context.line(`You're running tests on Node ${process.version}, but Zapier runs your code on ${LAMBDA_VERSION}. The version numbers must match. See https://zapier.github.io/zapier-platform-cli/index.html#requirements for more info.`);
+  if (!utils.correctVersion()) {
     process.exitCode = 1;
     return Promise.resolve();
   }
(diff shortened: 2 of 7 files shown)


src/constants.js
```diff
@@ -36,17 +37,18 @@ zzzzzzzzzzzzzzz          zzzzzzzzzzzzzzz
...
   BUILD_DIR,
   BUILD_PATH,
-  ART
+  CURRENT_APP_FILE,
+  DEBUG,
+  DEFINITION_PATH,
+  ENDPOINT,
+  LAMBDA_VERSION,
+  PLATFORM_PACKAGE,
+  STARTER_REPO
```
bcooksey: Thanks for cleaning these up!

src/commands/test.js
```diff
@@ -13,8 +13,7 @@ const test = (context) => {
     extraEnv.DETAILED_LOG_TO_STDOUT = 'true';
   }
 
-  if (process.version !== LAMBDA_VERSION) {
-    context.line(`You're running tests on Node ${process.version}, but Zapier runs your code on ${LAMBDA_VERSION}. The version numbers must match. See https://zapier.github.io/zapier-platform-cli/index.html#requirements for more info.`);
+  if (!utils.correctVersion()) {
```
bcooksey: Does this need to pass `context` in?

Also, what do you think about naming it `isCorrectNodeVersion`?

src/entry.js
```diff
@@ -22,6 +22,8 @@ module.exports = (argv) => {
     process.exit(1);
   }
 
+  require('update-notifier')({ pkg: require('../package.json') }).notify();
```
bcooksey: Is notifier smart enough to not shoot off warnings while we are developing the lib?

src/constants.js
xavdid: If you like this, there's more where that came from! #89

src/commands/test.js
xavdid: I figured that was useful for being able to use `context.line` instead of `console.log`, but I may be underestimating. 

Happy to rename, that's probably a good way to signal it just returns a bool. Also thought about moving `process.exit_status = 1` into the function itself, but wasn't sure if we wanted the side effect in there.

src/entry.js
xavdid: It's not, but it'll only pop it up once a day. Alternatively, it's easy to [opt out](https://github.com/yeoman/update-notifier#user-settings)

src/commands/test.js
bcooksey: I think how you designed it now is :100:, just a name to help communicate the intent of the function.

My question on context was that the new function assumes context exists and does `context.line`, so if we don't pass it, won't that blow up?

xavdid: ohhh you're super right. I thought the question is "is it really necessary to pass context into this function at all". I'll fix!

xavdid:
Thanks for the 👀!

xavdid:
Just for the package itself. We could configure it to check more (I think just run multiple copies), but the deps will update when the package itself updates.

package.json
```diff
@@ -39,7 +39,8 @@
     "read": "1.0.7",
     "string-length": "1.0.1",
     "through2": "2.0.1",
-    "tmp": "0.0.28"
+    "tmp": "0.0.28",
+    "update-notifier": "^2.2.0"
```
ibolmo: Just noticed this has the ^. Might want to freeze it to `"2.2.0"` in a different commit/PR.

xavdid: ah, nice call. Probably a good play, yeah.

xavdid: I'll probably go through the [deps](https://david-dm.org/zapier/zapier-platform-cli) and try to update as widely as I can before the 2.0 release.

PR #89, opened 2017-06-07
(minor) Alphabetize exports

xavdid:
Be more consistent in the way we list functions in modules.exports.

```js
modules.exports = {
  b,
  c,
  a
}
```

becomes

```js
modules.exports = {
  a,
  b,
  c,
}
```

Also grabbed my changes from #88, so if we're not going to merge that, we should clean this up (branches are hard).


xavdid:
Yeah, I started alphabetizing halfway through that other branch so I just committed them separately. Merge will look right once that other branch goes through.

src/commands/index.js
```diff
@@ -1,27 +1,27 @@
 module.exports = {
```
ibolmo: wonder if it's better to do this dynamically:

```js
module.exports = sortObjectKeys(module.exports); // see below
```

Since this is done in other places perhaps a utility function:

```js
const sortObjectKeys = (object) => (
  Object.keys(object).sort().reduce((sorted, key) => (sorted[key] = object[key], sorted), {})
);
```

xavdid: heh, so If we're not typing the modules out, their order doesn't matter. This PR is purely organizational for humans, so i'm not sure automating it gets us anywhere. 😁

PR #92, opened 2017-06-14
Fix minor typos in promote logs

dougshamoo:
Just noticed this while I was fixing some issues with our zapier application.


BrunoBernardino:
Thank you so much for this contribution @dougshamoo ! Unfortunately we're still confirming if/how can we accept contributions directly (legal stuff) and as such I won't be able to merge this PR. I'll close it and fix this through a PR of our own.

Thank you again for your time!

PR #96, opened 2017-06-19
Fix typo in session auth docs

bcooksey:


bcooksey:
Tried to do `npm run docs` and a lot changed, including reverts of docs saying we run 4.3.2 by default. @kola-er or @xavdid anything I need to do so building the docs works correctly?

xavdid:
@bcooksey I just tested on your branch and couldn't reproduce. Are you on node `v6.10.2` and have you run `npm install` on that version? 

`LAMBDA_VERSION` is set in `constants.js`, which is correct in your branch.

bcooksey:
Ah, nuking `node_modules` and `npm install` was the ticket, thanks @xavdid!

PR #102, opened 2017-07-04
require an exact core version in package.json

xavdid:
This will resolve the server not knowing what to do with core version `^2.0.0` as discussed [here](https://zapier.slack.com/archives/C151BFEJ3/p1498842593194328). `npm` lets you define a range in a lot of ways, so we wanted to check other characters, we could. 

Also have some small miscellaneous fixes, such as not showing a huge raw zip if `zapier build` is run with `--debug` (not useful, makes good logs hard to find).

### DIFF
--- a/src/entry.js
+++ b/src/entry.js
@@ -62,7 +62,7 @@ module.exports = (argv) => {
 
   if (!utils.isValidAppInstall(command)) {
     console.error(colors.red(
-      `Looks like your package.json is missing ${PLATFORM_PACKAGE} or you haven't run npm install yet!`
+      `Looks like your package.json is missing \`${PLATFORM_PACKAGE}\`, its version is a range, or you haven't run npm install yet!`
     ));
     process.exit(1);
   }
--- a/src/entry.js
+++ b/src/entry.js
@@ -62,7 +62,7 @@ module.exports = (argv) => {
 
   if (!utils.isValidAppInstall(command)) {
     console.error(colors.red(
-      `Looks like your package.json is missing \`${PLATFORM_PACKAGE}\`, its version is a range, or you haven't run npm install yet!`
+      `Looks like your package.json is missing \`${PLATFORM_PACKAGE}\`, its not restricted to a single version, or you haven't run npm install yet!`
     ));
     process.exit(1);
   }
--- a/src/utils/api.js
+++ b/src/utils/api.js
@@ -94,7 +94,10 @@ const callAPI = (route, options) => {
 
       if (constants.DEBUG || global.argOpts.debug) {
         console.log(`>> ${requestOptions.method} ${requestOptions.url}`);
-        if (requestOptions.body) { console.log(`>> ${requestOptions.body}`); }
+        if (requestOptions.body) {
+          let cleanedBody = _.assign({}, JSON.parse(requestOptions.body), {zip_file: 'raw zip removed in logs'});
+          console.log(`>> ${JSON.stringify(cleanedBody)}`);
+        }
         console.log(`<< ${res.status}`);
         console.log(`<< ${(text || '').substring(0, 2500)}\n`);
       } else if (hitError) {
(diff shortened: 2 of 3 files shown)


src/utils/api.js
```diff
@@ -94,7 +94,10 @@ const callAPI = (route, options) => {
 
       if (constants.DEBUG || global.argOpts.debug) {
         console.log(`>> ${requestOptions.method} ${requestOptions.url}`);
-        if (requestOptions.body) { console.log(`>> ${requestOptions.body}`); }
+        if (requestOptions.body) {
+          let cleanedBody = _.assign({}, JSON.parse(requestOptions.body), {zip_file: 'raw zip removed in logs'});
```
benzapier: Good idea!

src/utils/misc.js
```diff
@@ -86,13 +86,18 @@ const isValidNodeVersion = () => {
 };
 
 const isValidAppInstall = (command) => {
-  if (command === 'help' || command === 'init' || command === 'login' || command === 'apps' || command === 'convert') {
+  if (['help', 'init', 'login', 'apps', 'convert'].includes(command)) {
```
benzapier: Nice bit of cleanup there!

src/entry.js
```diff
@@ -62,7 +62,7 @@ module.exports = (argv) => {
 
   if (!utils.isValidAppInstall(command)) {
     console.error(colors.red(
-      `Looks like your package.json is missing ${PLATFORM_PACKAGE} or you haven't run npm install yet!`
+      `Looks like your package.json is missing \`${PLATFORM_PACKAGE}\`, its version is a range, or you haven't run npm install yet!`
```
kola-er: With `its version is a range`, I feel it implies the user should expect the version of the core dependency to be a range as opposed to what we really want: exact version. Or am I interpreting it wrongly?

src/utils/api.js
kola-er: Please, I don't get how the raw zip is removed. Is it an item with key `zip_file` in the outcome of `JSON.parse(requestOptions.body)`? If so, I got it then.

xavdid: That's correct! It comes out as a ton of text, which isn't helpful. Makes debugging really frustrating.

src/entry.js
xavdid: It seems clear to me. Each of the things in the list (missing package, version is a range, haven't installed) is a bad thing that's keeping the cli from working. It might be more clear if all 3 were negative, eg `version isn't a single version` or similar. Is that better?

src/utils/misc.js
```diff
@@ -86,13 +86,18 @@ const isValidNodeVersion = () => {
...
     return true;
   }
 
   let packageJson;
   try {
     packageJson = require(path.join(process.cwd(), 'package.json'));
+    const coreVersion = packageJson.dependencies[PLATFORM_PACKAGE];
+    // could check for a lot more, but this is probably enough: https://docs.npmjs.com/files/package.json#dependencies
+    if (!coreVersion || coreVersion.includes('^')) {
```
xavdid: I think it's a weird corner case that ben brought up [here](https://zapier.slack.com/archives/C151BFEJ3/p1498842593194328). Since that's a valid (and common) thing to do in `package.json`, we should flag it. The others are also technically possible but incredibly uncommon

xavdid:
@benzapier `init` uses whatever the template versions are set to (pinned, not range)

src/entry.js
kola-er: Yes, the negative one sounds clearer to me.

src/utils/misc.js
kola-er: Okay then, we are fine considering only `^`.

src/entry.js
xavdid: Sweet, I've updated. Thanks!

PR #105, opened 2017-07-04
Adding connectionLabel to some snippets as an example

BrunoBernardino:
Needs to be merged after https://github.com/zapier/zapier/pull/12448 and before https://github.com/zapier/zapier-platform-schema/pull/20

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,7 +419,8 @@ const authentication = {
   // "test" could also be a function
   test: {
     url: 'https://example.com/api/accounts/me.json'
-  }
+  },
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -419,7 +419,8 @@ const authentication = {
   // "test" could also be a function
   test: {
     url: 'https://example.com/api/accounts/me.json'
-  }
+  },
+  connectionLabel: '{{bundle.authData.username}}' // Can also be a function, check digest auth for an example
```
bcooksey: It's small, but what do you think about saying "check digest auth below" so folks know we mean to stay in this section? When I read this out of context, there were a number of ideas I had for what "check digest auth" could mean.

PR #106, opened 2017-07-05
Script enhancement for set-app-template-versions

kola-er:
This makes it possible to update the node and npm versions in the package.json file, the node version in the .nvmrc file, and the node_js version in the .travis.yml file of the app templates along with the version update for the zapier-platform-core in the package.json file.

Instead of the hardcoded value, the `LAMBDA_VERSION` constant is now gotten from a central source for versions. 

P.S. [Here](https://zapier.slack.com/archives/C151BFEJ3/p1498854268714539) is the Slack discussion on the need for it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,12 +38,10 @@
     "lodash": "4.17.4",
     "node-fetch": "1.7.1",
     "read": "1.0.7",
-    "semver": "5.3.0",
     "string-length": "1.0.1",
     "through2": "2.0.3",
     "tmp": "0.0.31",
-    "update-notifier": "2.2.0",
-    "yamljs": "0.3.0"
+    "update-notifier": "2.2.0"
   },
   "devDependencies": {
     "babel-cli": "6.24.1",
@@ -57,7 +55,9 @@
     "mocha": "3.4.2",
     "ngrok": "2.2.10",
     "node-watch": "0.5.4",
-    "should": "11.2.1"
+    "semver": "5.3.0",
+    "should": "11.2.1",
+    "yamljs": "0.3.0"
   },
   "bin": {
     "zapier": "zapier.js"
--- a/package.json
+++ b/package.json
@@ -38,10 +38,12 @@
     "lodash": "4.17.4",
     "node-fetch": "1.7.1",
     "read": "1.0.7",
+    "semver": "5.3.0",
     "string-length": "1.0.1",
     "through2": "2.0.3",
     "tmp": "0.0.31",
-    "update-notifier": "2.2.0"
+    "update-notifier": "2.2.0",
+    "yamljs": "0.3.0"
   },
   "devDependencies": {
     "babel-cli": "6.24.1",
--- a/scripts/set-app-template-versions.js
+++ b/scripts/set-app-template-versions.js
@@ -13,6 +13,8 @@ const yaml = require('yamljs');
 const childProcess = utils.promisifyAll(require('child_process'));
 
 const CLONE_URL_PREFIX = 'git@github.com:zapier/zapier-platform-example-app';
+const PACKAGES_NAMES = `node, npm, and zapier-platform-core`;
+let packagesVersions;
 
 const newCoreVersion = process.argv[2];
 if (!newCoreVersion) {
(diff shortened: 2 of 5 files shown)


kola-er:
I couldn't think of how to write test for the script so I manually tested against forks for [minimal](https://github.com/kola-er/zapier-platform-example-app-minimal/commit/e0cfe907d24d96597a832d99cb08a2a7c67dbf55) and [trigger](https://github.com/kola-er/zapier-platform-example-app-trigger/commit/992ae88bae29c2fbe038cfa87f6f084b7c2320ac) app templates. Click on the links to see the commits. Here is the terminal output:

package.json
```diff
@@ -38,10 +38,12 @@
     "lodash": "4.17.4",
     "node-fetch": "1.7.1",
     "read": "1.0.7",
+    "semver": "5.3.0",
     "string-length": "1.0.1",
     "through2": "2.0.3",
     "tmp": "0.0.31",
-    "update-notifier": "2.2.0"
+    "update-notifier": "2.2.0",
+    "yamljs": "0.3.0"
```
xavdid: `yamljs` and `semver` need to be in `devDependencies` instead of regular ones.

scripts/set-app-template-versions.js
```diff
@@ -36,28 +42,42 @@ const setVersion = (template, rootTmpDir) => {
...
 
-      if (packageJson.dependencies['zapier-platform-core'] === newVersion) {
+      const nvmrcFile = path.resolve(rootTmpDir, `${repoName}/.nvmrc`);
+      const nvmrcNodeVersion = fse.readFileSync(nvmrcFile, 'utf8').trim().substr(1); // strip off leading 'v'
+
+      const travisYamlFile = path.resolve(rootTmpDir, `${repoName}/.travis.yml`);
+      const travisYaml = yaml.load(travisYamlFile);
+
+      const nodeVersion = semver.Comparator(newVersions.nodeVersion).semver.version
+      if (packageJson.dependencies['zapier-platform-core'] === newVersions.coreVersion && packageJson.engines['node'] === newVersions.nodeVersion && packageJson.engines['npm'] === newVersions.npmVersion && nvmrcNodeVersion === nodeVersion && travisYaml.node_js[0] === nodeVersion) {
```
xavdid: should probably move this into a function (like `hasCurrentVersions`). Reusable, plus it'll be easy to put it on multiple lines to be more clear about what's being checked. 

```javascript
return (
  packageJson.dependencies['zapier-platform-core'] === newVersions.coreVersion &&
  packageJson.engines['node'] === newVersions.nodeVersion && 
  // ... 
)
```

scripts/set-app-template-versions.js
```diff
@@ -70,14 +90,14 @@ const setVersion = (template, rootTmpDir) => {
     })
     .then(result => {
       if (result === 'skip') {
-        console.log(`${template} is already set to ${newVersion}, skipping`);
+        console.log(`${template} is already set to ${newVersions.nodeVersion}, ${newVersions.npmVersion}, and ${newVersions.coreVersion} for node, npm, and zapier-platform-core respectively, skipping`);
```
xavdid: it'd be nice to see `${newVersions.nodeVersion}, ${newVersions.npmVersion}, and ${newVersions.coreVersion} for node, npm, and zapier-platform-core` in some sort of function since its used a lot. There's also some variation between the messages, so it may not be worth it.

src/version-store.js
```diff
@@ -0,0 +1,7 @@
+// Index of each item is equivalent to the major of zapier-platform-core version.
+// The next item will be for zapier-platform-core version of major 3.
+module.exports = [
+  {nodeVersion: '>=4.3.2', npmVersion: '>=2.14.12'},
+  {nodeVersion: '>=4.3.2', npmVersion: '>=2.14.12'},
+  {nodeVersion: '>=6.10.2', npmVersion: '>=3.10.10'}
```
xavdid: If we're requiring exact node versions, is it worth proving those instead of `>=`? npm has some wiggle room, so it's less of a big deal.

package.json
BrunoBernardino: Agreed.

scripts/set-app-template-versions.js
BrunoBernardino: Echoing David here.

scripts/set-app-template-versions.js
BrunoBernardino: I don't see a lot of value to that, honestly, but wouldn't be bad to dit.

src/version-store.js
BrunoBernardino: Yeah, I agree `nodeVersion` should be exact, and only `npmVersion` should be `>=`.

package.json
kola-er: Silly mistake. Thanks for catching it.

kola-er: Wait a minute... Those 2 packages are required in running the script and not in developing it.

kola-er:
I would love to look into the mocking of fetching the repos. Considering that those app templates need to be updated asap, I think the mock tests could be a separate PR. cc @BrunoBernardino

kola-er:
@xavdid @BrunoBernardino I think I've resolved the raised issues. Please, kindly review again. Thanks.

BrunoBernardino:
👍 ✅

PR #109, opened 2017-07-13
Allow deleting versions, improve zapier test and docs

### LINKED ISSUE #107: Allow skipping validate on zapier test
Just as we allow skipping on `zapier validate`, we should allow `--skip-validate` on `zapier test`.

BrunoBernardino:
Related to https://github.com/zapier/zapier/pull/12640, https://github.com/zapier/zapier-platform-cli/issues/107, https://trello.com/c/cMey6KY8/20-be-more-clear-in-the-cli-documentation-what-really-happens-with-zapier-migrate-and-zapier-deprecate, and https://trello.com/c/xY60Exu2/26-enable-zapier-logs-versionxxx

## Log

---

**Adding `zapier delete` command Making it possible to delete an app version, as long as it doesn't have any user/Zap.** (65328b3)

---

**Adding --skip-validate for `zapier test` This way it's possible to benefit from the environment setup and not have to wait for validate to always run. Fixes #107** (87aca19)

---

**Improving documentation on migration and deprecation** (3455347)

---

**Improving delete command** (b2fb8cb)

Moving delete and delete version to arguments, allow deleting a whole app.

Cleanup as per review suggestions.

Related to #109

---

**Fixing `zapier logs --version=1.0.0` throwing `zapier --version` instead** (270b050)

---

**Code cleanup for #109** (a603b18)

---

**Splitting the app and version deletion code for increased readibility.** (0a5cb5e)

Related to #109


src/commands/deprecate.js
```diff
@@ -32,7 +32,13 @@ deprecate.argOptsSpec = {};
 deprecate.help = 'Mark a non-production version of your app as deprecated, with removal by a certain date.';
 deprecate.example = 'zapier deprecate 1.0.0 2017-01-20';
 deprecate.docs = `\
```
ibolmo: For backticks you will not need the `\` to escape new lines.

src/commands/deprecate.js
```diff
@@ -32,7 +32,13 @@ deprecate.argOptsSpec = {};
...
-A utility to alert users of breaking changes that require the deprecation of an app version. Zapier will send emails warning users of the deprecation.
+A utility to alert users of breaking changes that require the deprecation of an app version.
+
+Use this when an app version will not be supported or start breaking at a known date.
+
+Zapier will send an email warning users of the deprecation once a date is set, they'll start seeing it as "Deprecated" in the UI, and once the deprecation date arrives, if the Zaps weren't updated, they'll be paused and the users will be emailed again explaining what happened.
+
+At that point it should be safe to delete that app version.
```
ibolmo: Which point? The point they inform, or on the deprecation date?

src/commands/migrate.js
```diff
@@ -35,7 +35,9 @@ Starts a migration to move users between different versions of your app. You may
 
 Only migrate users between non-breaking versions, use \`zapier deprecate\` if you have breaking changes!
 
-Migrations can take between 5-10 minutes, so be patient and check \`zapier history\` to track the status.
+Migrations can take between 5-10 minutes, so be patient and check \`zapier history\` to track the status
+
+This process should be transparent to users, as they aren't emailed about this update/migration, and if there are no breaking changes, the Zaps should continue to work as they were.
```
ibolmo: I'd say break this up, or remove the extra `if there are no breaking changes` since Line 36 already explains this. 

Maybe as simple as:

`Note: since a migration is only for non-breaking changes, users are not emailed about the update/migration.`

src/commands/test.js
```diff
@@ -21,7 +21,15 @@ const test = (context) => {
     return Promise.resolve();
   }
 
-  return validate(context)
+  let validated;
+
+  if (global.argOpts['skip-validate']) {
+    validated = Promise.resolve();
+  } else {
+    validated = validate(context);
```
ibolmo: Nitpick (feel free to ignore) use a ternary.

src/commands/delete.js
```diff
@@ -0,0 +1,48 @@
+const utils = require('../utils');
+
+const _delete = (context, version) => {
+  if (!version) {
+    context.line('Error: No version - provide a version like "1.0.0"...\n');
+    return Promise.resolve(true);
```
ibolmo: Should this be a `Promise.reject`? Not sure what the `true` would refer to.

ibolmo: I'm also seeing `throw new Error` in other commands.

src/commands/delete.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+  return utils.checkCredentials()
+    .then(() => utils.getLinkedApp())
+    .then((app) => {
+      context.line(`Preparing to delete version ${version} of your app "${app.title}".\n`);
```
ibolmo: I'm not sure of the extra `\n` I see the `context.line` is just `console.log`. That should already add a new line?

goodies/bash/_zapier
```diff
@@ -31,7 +31,7 @@ _zapier() {
                 COMPREPLY=( $( compgen -W "--dest --entry" -- $cur ) )
                ;;
             test)
-                COMPREPLY=( $( compgen -W "--debug --timeout" -- $cur ) )
+                COMPREPLY=( $( compgen -W "--debug --timeout --skip-validate" -- $cur ) )
```
ibolmo: I'm not sure I'd expect for test to skip validation. I might have missed the discussion, though.

src/commands/delete.js
```diff
@@ -0,0 +1,48 @@
...
+_delete.argsSpec = [
+  {name: 'version', example: '1.0.0', required: true, help: 'the version to delete'},
+];
+_delete.argOptsSpec = {};
+_delete.help = 'Delete a version of your app as long as it has no users/Zaps.';
+_delete.example = 'zapier delete 1.0.0';
+_delete.docs = `\
+A utility to allow deleting app versions that aren't used.
+
+> The app version needs to have no users/Zaps in order to be deleted.
```
ibolmo: small improvement: `The app version can only be deleted if it has no users or live Zaps.` or shorter: `The app version can only be deleted if it has no live Zaps.`

src/commands/delete.js
```diff
@@ -0,0 +1,48 @@
...
+
+**Arguments**
+
+${utils.argsFragment(_delete.argsSpec)}
+${utils.argOptsFragment(_delete.argOptsSpec)}
+
+${'```'}bash
+$ zapier delete 1.0.0
```
bcooksey: `delete` from a code perspective is solid, however, we should discuss the UX.

First up is whether we should be consistent within the CLI tool. For `invite` and `env`, we've chosen `--remove` as the way to nuke a specific piece of data. This introduces a different pattern. Granted, in those cases, there is a required argument, so the commands enforce what is getting removed. With `versions`, we'd have to allow an optional argument to specify which one. Still, I think that may be preferable to `zapier delete`.

A couple other considerations:

1) From the command alone, it's not clear what `delete` is going to allow you to remove. Apps? App Versions?
 Both? More stuff? Also, when we decide to allow deleting apps, what are we going to pick as the command if delete is already reserved for versions?
2) Related to one, should we go ahead and allow deleting apps as well? Right now you can nuke the versions, but then you'd have to go into the UI to nuke the app itself, which is a little inconsistent. I've heard devs ask in the public Slack channel to delete an test app they used to experiment, so there is a use case for deleting an entire app in one go.

bcooksey: fwiw, I wouldn't be opposed to going with a syntax like `zapier delete app` or `zapier delete version 1.0.0`.

BrunoBernardino: Great questions! I pondered doing `zapier versions 1.0.0 --remove` but since the process to add a version is to use `zapier push`, it seemed fitting to have a separate command to remove (unlike `zapier invite` or `zapier env` where we use that same command to add).

For 1) and 2) are you suggesting a `zapier delete-app` or `zapier delete-version 1.0.0`, for example? I'd definitely prefer `zapier delete app` and `zapier delete version 1.0.0` there.

goodies/bash/_zapier
BrunoBernardino: It's opt-in, meaning it won't skip validation by default, but you can make it skip it by passing `--skip-validate`.

src/commands/delete.js
bcooksey: Got it, I follow with the distinct command. In that case, I'm :+1: for `zapier delete app` and `zapier delete version 1.0.0` (or `zapier remove app`)

src/commands/delete.js
BrunoBernardino: It's to have an extra blank line for extra readability.

BrunoBernardino:
Alright! @bcooksey / @ibolmo please review again!

src/commands/delete.js
```diff
@@ -0,0 +1,53 @@
+const utils = require('../utils');
+
+const _delete = (context, appOrVersion, version) => {
+  if (appOrVersion === 'version' && !version) {
```
ibolmo: Just realized that the comparison might be better understood if put at the root scope of the call.

E.g. 
```js
const isDeletingVersion = appOrVersion === 'version';
if (isDeletingVersion && !version) {
...
```

src/commands/delete.js
```diff
@@ -0,0 +1,53 @@
...
+    const message = 'Error: No version - provide a version like "1.0.0"...';
+    return Promise.reject(new Error(message));
+  }
+  return utils.checkCredentials()
+    .then(() => utils.getLinkedApp())
+    .then((app) => {
+      const deletePrepMessage = (appOrVersion === 'version')
```
ibolmo: I'd also argue that this could be moved out and separated. E.g.

```js
return isDeletingVersion ? deleteVersion() : deleteApp();
```

I'm not as familiar with the code, but I would have expected an `App` model that would have these methods/actions. That would be a non-blocker, though, if it's not the norm.

src/commands/delete.js
```diff
@@ -0,0 +1,53 @@
...
+      return utils.callAPI(url, {
+        method: 'DELETE',
+      });
+    })
+    .then(() => {
+      utils.printDone();
+      context.line('  Deletion successful!\n');
+    });
+};
+_delete.argsSpec = [
```
ibolmo: Food for thought (non-blocker) maybe we would (in the future) create a `command` util function that does:

```js
module.exports = command({
     argsSpec: [...],
    ...
    example: '...',
    run(context, appOrVersion, version) {

    }
});
```

It would just do all of the manual `_delete.example = ` stuff and return the run function as the export.

src/commands/deprecate.js
```diff
@@ -2,8 +2,8 @@ const utils = require('../utils');
 
 const deprecate = (context, version, deprecation_date) => {
   if (!deprecation_date) {
```
ibolmo: Just realized this needs to be a camelCase: `deprecationDate`

src/commands/deprecate.js
```diff
@@ -2,8 +2,8 @@ const utils = require('../utils');
...
-    context.line('Error: No version or deprecation date - provide a version like "1.0.0" and "2017-01-20"...\n');
-    return Promise.resolve(true);
+    const message = 'Error: No version or deprecation date - provide a version like "1.0.0" and "2017-01-20"...';
+    return Promise.reject(new Error(message));
```
ibolmo: Could prob just inline since we don't have a max line length in JS.

src/entry.js
```diff
@@ -46,7 +46,7 @@ module.exports = (argv) => {
   // create the context, logs thread through this
   const context = utils.createContext({command, args, argOpts});
 
-  if (argOpts.version || argOpts.v) {
+  if (command === 'help' && (argOpts.version || argOpts.v)) {
```
ibolmo: Ah, this was added because in other commands (e.g. `zapier delete --version` would have exited early. 

`zapier --version` is not gone, though. Could we instead do `!command` or similar to try to keep the `zapier --version` working?

BrunoBernardino: Can you please confirm this for me again? `zapier --version` is working for me, since `help` becomes the default command if none is specified (look about 4-6 lines above).

I definitely agree `zapier --version` should work, as it's standard, like `zapier -v`.

src/commands/delete.js
BrunoBernardino: I think that might be a good idea for a refactor in the future, yes.

src/commands/delete.js
BrunoBernardino: Indeed that's not a pattern we have here, but in a future refactor it would make things clearer indeed.

BrunoBernardino:
Please check again @ibolmo, and thank you so much for the thorough review!

src/entry.js
ibolmo: Ah gotcha. All good then 👍

src/commands/delete.js
```diff
@@ -0,0 +1,54 @@
...
+    return Promise.reject(new Error(message));
+  }
+  return utils.checkCredentials()
+    .then(() => utils.getLinkedApp())
+    .then((app) => {
+      const deletePrepMessage = isDeletingVersion
+        ? `Preparing to delete version ${version} of your app "${app.title}".\n`
+        : 'Preparing to delete all versions of your app "${app.title}".\n';
+      context.line(deletePrepMessage);
+      const url = isDeletingVersion ? `/apps/${app.id}/versions/${version}` : `/apps/${app.id}`;
```
ibolmo: Sorry, I wasn't explicit. Splitting up the code here would be helpful for readability/maintainability. The moving of funcs to a Model could be done at a later time. Splitting this up would remove all the ternaries.

BrunoBernardino: Got it.

BrunoBernardino:
Awesome points @ibolmo ! I've updated the code, thanks!

PR #118, opened 2017-08-03
Adding warning when a build is uploaded with an outdated core version.

BrunoBernardino:
Related to https://github.com/zapier/zapier/pull/12836 and needs it shipped before this can ship.

Sample screenshot:

cli-needs-update

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "lodash": "4.17.4",
     "node-fetch": "1.7.1",
     "read": "1.0.7",
+    "semver": "5.3.0",
     "string-length": "1.0.1",
     "through2": "2.0.3",
     "tmp": "0.0.31",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -55,7 +56,6 @@
     "mocha": "3.4.2",
     "ngrok": "2.2.10",
     "node-watch": "0.5.4",
-    "semver": "5.3.0",
```
kola-er: @BrunoBernardino Please, check @xavdid's first comment on `package.json` in [this PR](https://github.com/zapier/zapier-platform-cli/pull/106) as regards making `semver` a dev dependency.

BrunoBernardino: Sure thing. That comment doesn't apply here because this will run by someone who's installed `zapier-platform-cli` via `npm` (on `zapier upload` and `zapier push`).

`devDependencies` are only installed on an explicit `npm install` inside a `zapier-platform-cli` folder, whereas `dependencies` are installed with a `npm install -g zapier-platform-cli` (`-g` being irrelevant here). Does that make sense?

PR #122, opened 2017-08-18
Use path.join to form windows-compatible path

FokkeZB:
Not sure if this causes:
https://secure.helpscout.net/conversation/415601400/628033/?folderId=20363#thread-1123745601

But still good practice not to use os-specific path separator.

### DIFF
--- a/src/utils/build.js
+++ b/src/utils/build.js
@@ -226,7 +226,7 @@ const build = (zipPath, wdir) => {
       printDone();
       printStarting('Applying entry point file');
       // TODO: should this routine for include exist elsewhere?
-      return readFile(`${tmpDir}/node_modules/${constants.PLATFORM_PACKAGE}/include/zapierwrapper.js`)
+      return readFile(path.join(tmpDir, 'node_modules', constants.PLATFORM_PACKAGE, 'include', 'zapierwrapper.js'))
         .then(zapierWrapperBuf => writeFile(`${tmpDir}/zapierwrapper.js`, zapierWrapperBuf.toString()));
     })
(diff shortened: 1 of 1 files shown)


FokkeZB:
It also looks like this check won't work on Windows:

https://github.com/zapier/zapier-platform-cli/blob/6156a43b91851a3ec9d952dcc2193541394419d1/src/utils/build.js#L116

FokkeZB:
@BrunoBernardino what do you think about https://github.com/zapier/zapier-platform-cli/blob/6156a43b91851a3ec9d952dcc2193541394419d1/src/utils/build.js#L116? Should I also update that to be cross-platform?

BrunoBernardino:
I think that makes sense!

src/utils/build.js
```diff
@@ -113,7 +113,7 @@ const forceIncludeDumbPath = (filePath/*, smartPaths*/) => {
   // we include smartPaths just incase you want to check the inclusion of some library
   return (
     (filePath.endsWith('package.json') || filePath.endsWith('definition.json'))
-    || filePath.match(/aws-sdk\/apis\/.*\.json/)
+    || filePath.match(path.sep === '\\' ? /aws-sdk\\apis\\.*\.json/ : /aws-sdk\/apis\/.*\.json/)
```
FokkeZB: @BrunoBernardino since Travis doesn't support Windows (yet) it's hard to add a test, but I've tested this logic in Parallels:

BrunoBernardino: Sounds good enough to me!

PR #130, opened 2017-08-30
Adding style checks to `zapier build`

### LINKED ISSUE #103: `zapier build` doesn't directly call `zapier validate`
I noticed this when making a validation-related fix. `zapier build` says it validates your project. It uses `_appCommandZapierWrapper`, which in turn wraps core's `schemaTools.validateApp`([source](https://github.com/zapier/zapier-platform-core/blob/master/src/tools/schema.js#L121-L125)). 

This is mostly fine, but that means any extra validation that happens as part of the `validate` command (in this case, running style checks), only happens when `validate` is called directly, but not during `build`. 

Resolution would be to wrap the `validate` command in `cli` rather than going straight to `core`.

BrunoBernardino:
Fixes #103

Initially I tried tackling it by _really_ calling `zapier validate` but that command runs a lot more things that don't belong in the `build` or `push` context.

After that I tried to not make duplicate calls or generate the definition file more than once, which required moving some build tasks around.

Also added an option to show any write file error with --debug.

I tried adding tests but since this is for the whole `zapier build` or `zapier push`, it's very difficult. I ended up just making a lot of tests with apps, locally.

[Relevant Trello Card](https://trello.com/c/21gffyuk/8-call-zapier-validate-from-zapier-build).

### What this changes

- [x] `zapier build` now errors if there are failing style checks
- [x] `zapier push` now errors if there are failing style checks in the `build` part, instead of the `upload` part (sooner)

### Screenshots

Failing `push` because of validation now, on the `build` level.
screen shot 2017-08-30 at 12 43 50

Failing `build` for a non-style-check-related failure.
screen shot 2017-08-30 at 12 44 00

Passing `push`
screen shot 2017-08-30 at 12 43 45

### DIFF
--- a/src/utils/build.js
+++ b/src/utils/build.js
@@ -29,11 +29,12 @@ const {
 const {
   checkCredentials,
   upload,
+  callAPI,
 } = require('./api');
 
 const {
   runCommand,
-  isWindows
+  isWindows,
 } = require('./misc');
 
 const stripPath = (cwd, filePath) => filePath.split(cwd).pop();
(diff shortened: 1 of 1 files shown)


BrunoBernardino:
Great points! Thanks for looking into this. Let me see if I can address all of your comments:

1. The main thing to note is the `validate` and `definition` "core" commands are running from different locations (locally in `zapier validate` and from the tmp dir for the build in `zapier build`), so they're not the exact same.

2. The api call seemed too small to abstract into a separate function, but I'm happy to do it if you think it's worthwhile. The `validate` and `definition` are already the most abstracted they can be, I think.

3. The main "extra" things the `zapier validate` does is related to output (it outputs much more information on success and errors, which isn't consistent within the context of how `zapier build`, `zapier upload`, and `zapier push` behave). Running the command itself makes it harder to have both command outputs behave consistently. Also, I don't think we want to allow skipping style checks when building, but it's fine when validating.

4. I completely understand the concern with the commands not being in sync, and that's why I tried calling `zapier validate` initially directly. Unless we build a specific `zapier validate --silent` that isn't _really_ silent but only outputs something like OK/NOTOK, it's not possible to accomplish that. While I don't think the code's at a point where we need to consider that the best possibility (seems like the duplication isn't that great, all in all), I think we'll eventually get there in the future, if the complexity keeps growing.

I hope that makes sense? Let me know your thoughts.

xavdid:
Yep, that all makes sense!

When I had first looked, I thought the function could be something like `validateSchemaAndStyle` that takes an app definition, does the api calls and processes them. Looking back now I'm not sure they're similar enough for it to be worth it. 

I say ship it!

PR #140, opened 2017-09-14
Add Onedrive as a valid template

bcooksey:
We just published https://github.com/zapier/zapier-platform-example-app-onedrive

This makes it so that `zapier init --template=onedrive ./` works!

### DIFF
--- a/src/app-templates.js
+++ b/src/app-templates.js
@@ -14,4 +14,5 @@ module.exports = [
   'babel',
   'rest-hooks',
   'files',
+  'onedrive'
 ];


bcooksey:
I see your point. I'll check on the validation scripts to see if that will be a pain. If so, we can leave it out.

My thought on adding it to the list is that the `init` are to inspire you, showing you what an app can do. I think of `scaffold` as the boilerplate handler, which I agree, we wouldn't want to have dump an entire app into your project.

Either way, your comment about adding to the README is also good.

xavdid:
@BrunoBernardino @bcooksey I realize this is old, but it's nice to keep the PR backglog clean. 

Seems like the consensus here is that though Onedrive is a good app, it's not a template that you'd use to start your app (plus it would mess with being able to run tests on all templates), so it seems like we're safe to close?

BrunoBernardino:
I'd say the only action here is to link the app/repo in the readme as a full-fledged example.

bcooksey:
Sounds like a plan. The CLI docs direct you to [this page](https://github.com/zapier/zapier-platform-cli/wiki/Example-Apps) which already has onedrive, so think we can close this

PR #144, opened 2017-09-25
Documentation Fixes

casshill:
Some code examples in the documentation were showing "helpText" instead of "description" in the display object, which doesn't match up with the schema documentation.


BrunoBernardino:
Thank you so much for this @casshill !

PR #145, opened 2017-09-25
Update z.request's body option for sugar

### LINKED ISSUE #143: Why does `z.request` work with an object for `body` ?
At https://github.com/zapier/zapier-platform-cli#http-request-options we repeat ​https://www.npmjs.com/package/node-fetch#options which says `body` "can be a string, buffer, or readable stream":

Yet in examples like https://github.com/zapier/zapier-platform-example-app-oauth2/blob/master/authentication.js#L2-L13 we set it to an object.

While https://github.com/bitinn/node-fetch/blob/master/src/body.js#L48 would convert that into a useless `[Object object]` string.

So the question is; why does this work? And assume it works, we should document how/why.

/cc @BrunoBernardino

FokkeZB:
Fixes #143

### DIFF
--- a/README.md
+++ b/README.md
@@ -1395,7 +1395,7 @@ Shorthand requests and manual `z.request([url], options)` calls support the foll
 * `method`: HTTP method, default is `GET`.
 * `headers`: request headers object, format `{'header-key': 'header-value'}`.
 * `params`: URL query params object, format `{'query-key': 'query-value'}`.
-* `body`: request body, can be a string, buffer, or readable stream. Default is `null`.
+* `body`: request body, can be a string, buffer, readable stream or plain object. When it is an object/array and the `Content-Type` header is `application/x-www-form-urlencoded` the body will be transformed to a query string, otherwise we'll set the header to `application/json; charset=utf-8` and JSON encode the body. Default is `null`.
 * `json`: shortcut object/array/etc. you want to JSON encode into body. Default is `null`.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1395,7 +1395,7 @@ Shorthand requests and manual `z.request([url], options)` calls support the foll
 * `method`: HTTP method, default is `GET`.
 * `headers`: request headers object, format `{'header-key': 'header-value'}`.
 * `params`: URL query params object, format `{'query-key': 'query-value'}`.
-* `body`: request body, can be a string, buffer, or readable stream. Default is `null`.
+* `body`: request body, can be a string, buffer, readable stream or plain object. When it is an object/array and the `Content-Type` header is `application/x-www-form-urlencoded` the body will be transformed to a query string, otherwise we'll set the header to `application/json; charset=utf-8` and JSON encode the body. Default is `null`.
```
bradleyayers: You might want to update the example below too (line 1416):

```
   // only provide body, json or form...
   body: '{"hello":"world"}', // or 'hello=world'
```

> the body will be transformed to a query string

This may be misinterpreted as meaning the URL of the request will be modified to include the body as the query string.

BrunoBernardino:
@FokkeZB you should be making these changes in `README-source.md` so they're not lost on the next `npm run docs` or release.

FokkeZB:
I keep forgetting that! Done!

PR #153, opened 2017-10-05
Add header to docs

mjyoung:
This is a WIP, but opening now in case anyone has thoughts on how I'm including the header HTML and CSS here.


mjyoung:
@BrunoBernardino When you get a chance can you provide some feedback on this WIP PR? Is there a better way for me to insert the header HTML and CSS?

docs/docs.css
```diff
@@ -0,0 +1,62 @@
...
+
+.docs-header__main {
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
+  padding: 1em 1.25em;
+}
+
+.docs-header__logo {
+  background: url(https://cdn.zapier.com/static/1DZPa0/images/frontend/onboarding/zapier-small-orange-logo.png)
```
BrunoBernardino: Q: Will this URL eventually become invalid after a new deploy?

docs/template.jst
```diff
@@ -15,13 +15,34 @@
 
     <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css">
     <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.5.0/styles/agate.min.css">
+    <link href="https://fonts.googleapis.com/css?family=Open+Sans" rel="stylesheet">
```
BrunoBernardino: I'd just switch from `https://` to `//` for consistency with the other `<link>`s.

docs/docs.css
mjyoung: That's a very good point @BrunoBernardino. Hmmm. Do we have a more permanent URL for common assets like these?

BrunoBernardino: Replied on Slack.

mjyoung:
@BrunoBernardino please review.

image

mjyoung:
@BrunoBernardino I've fixed the scroll position issue with the table of contents. I was unable to replicate the other bugs in 1 and 2 above in Chrome (stable and beta) .

BrunoBernardino:
Alright, I've tested locally and it's all working fine (the other bugs were because I opened as `file://`, but serving in an HTTP server works fine — the fonts were missing on `file://`)! Thanks @mjyoung

PR #154, opened 2017-10-07
fix: exit code 1 for only for errors

ifedapoolarewaju:
Currently when `zapier test` is run, it terminates with exit code 1, when just warnings are returned during style check validation. Warnings are not critical IMHO. This behaviour breaks builds indiscriminately, even when the warning doesn't necessarily apply to the project.

A current work around is to run with `--skip-validate`, but then this takes away other benefits of running the validation, like identifying actual errors for example.

This PR changes the behavior to only terminate with exit code 1 for actual errors and not warnings. :)

### DIFF
--- a/src/commands/validate.js
+++ b/src/commands/validate.js
@@ -83,7 +83,8 @@ const validate = (context) => {
         ['Link', 'link']
       ], ifEmpty, true);
 
-      if (styleErrors.length) {
+      // exit code 1 only for errors and not warnings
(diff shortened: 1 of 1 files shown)


BrunoBernardino:
@ifedapoolarewaju can you please sign the CLA so we can merge this?

ifedapoolarewaju:
@BrunoBernardino done! Thank you!

PR #155, opened 2017-10-09
DocFix: Errors are under z.errors

FokkeZB:
Readme examples had `throw new HalterError` which should be `throw new z.errors.HaltedError`.

### DIFF
--- a/docs/cli.html
+++ b/docs/cli.html
@@ -754,13 +754,12 @@ $ &#x2502; <span class="hljs-built_in">logout</span>      &#x2502; zapier <span 
   <div class="row-height">
     <div class="col-md-5 col-sm-12 col-height  docs-primary">
       <blockquote>
-<p>Manage the invitees/testers on your project. Can optionally specify a version or --remove.</p>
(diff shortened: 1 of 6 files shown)


docs/cli.html
```diff
@@ -754,13 +754,12 @@ <h2 id="invite">invite</h2>
   <div class="row-height">
     <div class="col-md-5 col-sm-12 col-height  docs-primary">
       <blockquote>
-<p>Manage the invitees/testers on your project. Can optionally specify a version or --remove.</p>
-</blockquote><p>  <strong>Usage:</strong> <code>zapier invite [user@example.com] [1.0.0]</code></p><p>Invite any user registered on Zapier to test your app. Commonly, this is useful for teammates, contractors, or other team members who might want to test, QA, or view your app versions. If you&apos;d like to provide full admin access, try <code>zapier collaborate</code>.</p><blockquote>
+<p>Manage the invitees/testers on your project. Can optionally --remove.</p>
```
FokkeZB: @BrunoBernardino it looks like documentation on invite versions were added without committing the change to `README-source.md`

FokkeZB: Found it - since CLI docs are generated from compiled `src` the `npm run build` should always be ran before `npm run docs`. Changed that in `package.json`.

Issue #139, opened 2017-09-13
The command `zapier build` omits non-code data files from `build.zip`


benzapier:
The command `zapier build` starts with the `index.js` file, and adds all referenced files/dependencies/modules into the final `build.zip`

Thus, any non-code "important data/config" files needed by your integration (or modules) will be missing when the code is pushed up to Zapier's systems - such as any/all `json`, `xml`, `yml`, or other text/data file(s).  Node modules like `xml2js` also trip over this issue, because it requires/needs a configuration file that's contained within its `node_modules` directory structure.

The current workaround is to make a script on your local dev environment, that performs the following steps:
  * `zapier build`  - creates the `build.zip` file
  * Shell scripting commands to `unzip` the `build.zip` into a temporary directory
  * Shell scripting commands that copy over the necessary data/config files into that temporary directory
  * Shell scripting commands to `zip` the files in that temporary directory, and create a new version of `build.zip`
  * `zapier upload` to send that updated `build.zip` to the Zapier system.

The most likely solution will be adding some kind of Zapier-app-config file, which includes the paths/names of any additional (non-code) files to be included when building the zip file.  Ideally, this should support file globbing, but may run into limitations on Windows file systems.

BrunoBernardino:
I've tried a solution at #158 because I needed to include a file in the checks (for the interpreter), but was hitting a problem and I didn't really need that solution, so I just left it there and you're welcome to look into it, or I will sometime in the future.

BrunoBernardino:
Tracking in linked PR.

PR #158, opened 2017-10-11
Allow to include a regexp of paths for the build file

BrunoBernardino:
This allows creating a `"includeInBuild"` property in `.zapierapprc` which will include those paths in the `build.zip` file.

[Relevant Trello Card](https://trello.com/c/BCE6mlzy/124-cli-add-include-flag-to-zapier-build-command).

### What this changes

- [x] Allows adding a `includeInBuild` property in `.zapieraprc`, reading from it while building the app.
- [x] Adds documentation and examples for how to use it

### Log

---

**Allow to include a regexp of paths for the build file** (9cc0d82)

This is an attempt to solve #139. While it doesn't work (the files in `--include-paths=` get added, but `definition.json` is missing, though it was written), I'm leaving it for someone to work in the future.

---

**Rework to allow "includeInBuild" property in `.zapierapprc`.** (4c3e655)

Tested and working. Didn't add tests as it was necessary to generate and load a .zapierapprc file, and also expose a few more methods.

---

**Break loop as soon as it's no longer needed** (b7cabc2)

Related to #158


BrunoBernardino:
@benzapier you might want to look at this and test it out, as I believe you needed this.

src/utils/build.js
```diff
@@ -110,13 +111,21 @@ const listFiles = (dir) => {
...
 };
 
-const forceIncludeDumbPath = (filePath/*, smartPaths*/) => {
-  // we include smartPaths just incase you want to check the inclusion of some library
+const forceIncludeDumbPath = (appConfig, filePath) => {
+  let matchesConfigInclude = false;
+  const configIncludePaths = _.get(appConfig, 'includeInBuild', []);
+  _.each(configIncludePaths, (includePath) => {
+    if (filePath.match(RegExp(includePath, 'i'))) {
+      matchesConfigInclude = true;
```
eliangcs: Should probably to add `return false;` here to break the loop.

BrunoBernardino: Nice catch! I didn't do it initially because this should never be a long list, but there's no need to unnecessarily run through the loop anyway.

BrunoBernardino: Unfortunately, turns out `return false;` there isn't stopping the iterator. You can test with:

```js
_.each(['a', 'b', 'c'], (letter) => {
    console.log(letter);
    if (letter === 'b') {
      return false;
    }
  });
```

BrunoBernardino: Code updated!

PR #159, opened 2017-10-12
Fix noun in `zapier convert`

eliangcs:
Fixes https://github.com/zapier/zapier-platform-app-converted-google-maps/blob/32892cf9526b20ac087dd326230bbf2948a2247c/triggers/place_types.js#L25.

### DIFF
--- a/src/utils/convert.js
+++ b/src/utils/convert.js
@@ -167,11 +167,13 @@ const renderStep = (type, definition, key) => {
     quote(definition.url) + ',' :
     `'http://example.com/api/${key}.json', // TODO this is just an example`;
 
+  const noun = definition.noun || _.capitalize(key);
+
   const templateContext = {
     KEY: snakeCase(key),
     CAMEL: camelCase(key),
-    NOUN: definition.noun,
-    LOWER_NOUN: key.toLowerCase(),
+    NOUN: noun,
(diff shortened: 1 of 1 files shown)


src/utils/convert.js
```diff
@@ -170,7 +170,7 @@ const renderStep = (type, definition, key) => {
   const templateContext = {
     KEY: snakeCase(key),
     CAMEL: camelCase(key),
-    NOUN: _.capitalize(key),
```
BrunoBernardino: We should probably add a default in case noun doesn't exist (it's optional in WB). I think `_.capitalize(key)` is a decent enough default.

src/utils/convert.js
```diff
@@ -170,7 +170,7 @@ const renderStep = (type, definition, key) => {
...
+    NOUN: definition.noun,
     LOWER_NOUN: key.toLowerCase(),
```
BrunoBernardino: Is this used? If so, might want to apply a similar logic as above.

src/utils/convert.js
eliangcs: Fixed in 53cd6d3.

src/utils/convert.js
eliangcs: Also fixed in 53cd6d3. Thanks :)

PR #168, opened 2017-10-26
Skip Style Checks on Invalid apps

bcooksey:
## Description

If a dev has an app that has major schema validation errors (but still parses as JSON), the `zapier validate` command will POST that to zapier.com to run the style checks. In some cases, this can cause the server to throw a 500. That is being fixed, and will soon correctly return a 403.

However, there really isn't a point in making that call at all if we know locally we are invalid. So, this PR updates the validate command to skip style checks on invalid apps.

## TODO

- [x] Skip style check API call on invalid app
- [x] Do not print success message if style checks are skipped due to invalid app
- [x] Do not print success if `-without-styles` flag is passed.

### DIFF
--- a/src/commands/validate.js
+++ b/src/commands/validate.js
@@ -72,6 +72,10 @@ const validate = (context) => {
       }
     })
     .then((styleResult) => {
+      if (global.argOpts['without-style']) {
+        return;
+      }
+
       // process errors
       let styleErrors = condenseIssues(styleResult);
       const ifEmpty = colors.grey('No style errors found during validation routine.');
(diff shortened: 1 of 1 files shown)


bcooksey:
I should mention that the `process.exitCode` may not be the cleanest solution now that we are skipping stuff. As an alternative we could use `process.exit()`, though it is nice to flow through the whole chain so that each callback can be responsible for saying messages like "Skipping because", vs the point calling `process.exit()` having to be aware of what all other things are skipped. Maybe a `throw` and `.catch` could solve that.

PR #170, opened 2017-10-31
zapier convert: Add tests ensuring default help text

eliangcs:
We already give users a default description/help text for a step when `zapier convert` can't find it in their app definition. This PR adds some tests to ensure that.

### DIFF
--- a/src/tests/utils/convert.js
+++ b/src/tests/utils/convert.js
@@ -140,6 +140,44 @@ describe('convert render functions', () => {
     });
   });
 
+  describe('render step', () => {
+    it('should fill with trigger default description', () => {
+      const v2Def = {
+        key: 'exampleStep',
+        noun: 'Example',
+        label: 'Example Step',
+      };
+
+      return convert.renderStep('trigger', v2Def).then(content => {
+        content.should.containEql("description: 'Triggers on a new example.'");
+      });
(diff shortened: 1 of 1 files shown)


src/tests/utils/convert.js
```diff
@@ -140,6 +140,44 @@ describe('convert render functions', () => {
     });
   });
 
+  describe('render step', () => {
+    it('should fill with trigger default description', () => {
```
BrunoBernardino: I believe you're missing a few `done()` calls here, since the tests are async.

BrunoBernardino:
I've noticed a problem, but I can't "unapprove" my previous approval, only create a new "request changes", but I don't think it's necessary.

src/tests/utils/convert.js
eliangcs: By [returning the Promise](https://github.com/zapier/zapier-platform-cli/pull/170/files#diff-3b1a6bd6d07ddfd7a243e89ee5d1ce92R151), Mocha will handle the asynchronous code. I confirmed it by messing around the assertion and it did break the tests.

BrunoBernardino: Oh, I wasn't aware of that! Thanks! Feel free to merge, then.

PR #171, opened 2017-10-31
zapier convert: Fix test code with Promises

eliangcs:
A couple of asynchronous tests in `convert.js` don't get run when you do `npm run test`. For some reasons the `done()` used originally doesn't work. This PR fixes it by returning the Promise for Mocha to pick up. See https://mochajs.org/#working-with-promises for more detail.

### DIFF
--- a/src/tests/utils/convert.js
+++ b/src/tests/utils/convert.js
@@ -40,7 +40,7 @@ describe('convert render functions', () => {
   });
 
   describe('authentication', () => {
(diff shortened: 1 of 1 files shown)


BrunoBernardino:
Nice catch! I'm fixing these and adding more as part of https://github.com/zapier/zapier-platform-cli/pull/167 so I think we can close this PR without merging.

PR #172, opened 2017-10-31
Better zapier convert support for triggers

BrunoBernardino:
This add support for converting triggers that use pre/post/full/pre+post scripting methods.

### DIFF
--- a/scaffold/convert/create.template.js
+++ b/scaffold/convert/create.template.js
@@ -1,10 +1,9 @@
 // "Create" stub created by 'zapier convert'. This is just a stub - you will need to edit!
 
-// create a particular <%= LOWER_NOUN %> by name
-const create<%= CAMEL %> = (z, bundle) => {
+const makeRequest = (z, bundle) => {
   const responsePromise = z.request({
     method: 'POST',
-    url: <%= URL %>
+    url: '<%= URL %>',
     data: JSON.stringify({
       EXAMPLE: bundle.inputData.EXAMPLE
     })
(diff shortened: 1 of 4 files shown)


BrunoBernardino:
To better test the "conversion", I'll work on building a script in a different PR which will get a few apps, convert them, and run `zapier tests` (like we do for checking the example apps).

src/utils/convert.js
```diff
@@ -212,18 +212,85 @@ const renderAuth = (definition) => {
...
+    } else if (position === 'post') {
+      methodSuffix = '_post_search';
+    } else {
+      methodSuffix = '_search';
+    }
+  }
+
+  const methodName = `${key}${methodSuffix}`;
+
+  return (js.indexOf(methodName) !== -1);
```
eliangcs: Maybe we want to strip out the comments from `js` first, but we can do it in a separate PR.

BrunoBernardino: Yes, great idea!

scaffold/convert/create.template.js
```diff
@@ -1,10 +1,9 @@
 // "Create" stub created by 'zapier convert'. This is just a stub - you will need to edit!
 
-// create a particular <%= LOWER_NOUN %> by name
-const create<%= CAMEL %> = (z, bundle) => {
+const makeRequest = (z, bundle) => {
```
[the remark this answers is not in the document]
BrunoBernardino: I'll do that, makes sense!

PR #177, opened 2017-11-06
Add option to filter tests

FokkeZB:
Passes on `--grep` to mocha so that you can filter tests to run.

### DIFF
--- a/src/commands/test.js
+++ b/src/commands/test.js
@@ -33,14 +33,16 @@ const test = (context) => {
       const env = _.extend({}, process.env, extraEnv);
       const commands = ['run', '--silent', 'test'];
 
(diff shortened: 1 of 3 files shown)


src/commands/test.js
```diff
@@ -38,6 +38,11 @@ const test = (context) => {
         commands.push(`--timeout=${global.argOpts.timeout}`);
       }
 
+      if (global.argOpts.grep) {
+        commands.push('--');
```
BrunoBernardino: If `global.argOpts.timeout` is also set, I believe this will fail, right? If so, please add a condition here, otherwise, please ignore my comment.

FokkeZB: Stuuuuupid mistake, fixed.

PR #178, opened 2017-11-07
zapier convert: Generate basic test code for a trigger/create/search

eliangcs:


scaffold/convert/create-test.template.js
```diff
@@ -0,0 +1,30 @@
+require('should');
+
+const zapier = require('zapier-platform-core');
+
+const App = require('../../index');
+const appTester = zapier.createAppTester(App);
+
+const utils = require('../utils');
```
BrunoBernardino: I don't believe this is necessary.

scaffold/convert/create-test.template.js
```diff
@@ -0,0 +1,30 @@
...
+
+describe('Creates - <%= KEY %>', () => {
+  zapier.tools.env.inject();
+
+  it('should create an object', (done) => {
+    const bundle = {
+      authData: utils.getAuthData(),
```
BrunoBernardino: I'd try to create some method above (that would differ depending on the app's auth type) and call it here instead.

scaffold/convert/search-test.template.js
```diff
@@ -0,0 +1,28 @@
+require('should');
+
+const zapier = require('zapier-platform-core');
+
+const App = require('../../index');
+const appTester = zapier.createAppTester(App);
+
+const utils = require('../utils');
```
BrunoBernardino: Same as above.

scaffold/convert/search-test.template.js
```diff
@@ -0,0 +1,28 @@
...
+
+describe('Searches - <%= KEY %>', () => {
+  zapier.tools.env.inject();
+
+  it('should get an object', (done) => {
+    const bundle = {
+      authData: utils.getAuthData(),
```
BrunoBernardino: Same as above.

scaffold/convert/create-test.template.js
```diff
@@ -0,0 +1,30 @@
...
+      inputData: {
+        name: 'Testing',
+      },
+    };
+
+    appTester(App.creates.<%= KEY %>.operation.perform, bundle)
```
BrunoBernardino: Is there a possibility of `KEY` having a `-`? If so, we may prefer doing `['<%= KEY %>']` here.

scaffold/convert/search-test.template.js
```diff
@@ -0,0 +1,28 @@
...
+    };
+
+    appTester(App.searches.<%= KEY %>.operation.perform, bundle)
```
BrunoBernardino: Same as above.

scaffold/convert/trigger-test.template.js
```diff
@@ -0,0 +1,26 @@
+require('should');
+
+const zapier = require('zapier-platform-core');
+
+const App = require('../../index');
+const appTester = zapier.createAppTester(App);
+
+const utils = require('../utils');
```
BrunoBernardino: Same as above.

scaffold/convert/trigger-test.template.js
```diff
@@ -0,0 +1,26 @@
...
+
+describe('Triggers - <%= KEY %>', () => {
+  zapier.tools.env.inject();
+
+  it('should get an array', (done) => {
+    const bundle = {
+      authData: utils.getAuthData(),
```
BrunoBernardino: Same as above.

scaffold/convert/trigger-test.template.js
```diff
@@ -0,0 +1,26 @@
...
+    };
+
+    appTester(App.triggers.<%= KEY %>.operation.perform, bundle)
```
BrunoBernardino: Same as above.

scaffold/convert/create-test.template.js
eliangcs: Good catch. Fixed in 0f46d2b.

scaffold/convert/create-test.template.js
eliangcs: Fixed all the `authData` issues in bf3c107.

eliangcs:
@BrunoBernardino thanks for the review! I have made some fixes based on your feedback. Have a look when you got a chance. Thanks!

PR #179, opened 2017-11-10
zapier convert: Handle empty and multi-line field help text

eliangcs:
Fixes two issues with field help text:

1. Don't pad help text if it's too short. Keep empty help text empty.
2. Escape line breaks in help text so that it doesn't break the template.

### DIFF
--- a/src/tests/utils/convert.js
+++ b/src/tests/utils/convert.js
@@ -39,15 +39,15 @@ describe('convert render functions', () => {
       });
     });
 
-    it('should pad help text that is too short', () => {
+    it('should keep empty help text empty', () => {
       const wbKey = 'test_field';
       const wbDef = {
-        help_text: 'too short'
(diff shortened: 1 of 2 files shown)


src/tests/utils/convert.js
```diff
@@ -97,14 +109,14 @@ describe('convert render functions', () => {
                 type: 'string',
                 required: true,
                 label: 'Username',
-                helpText: '(help text must be at least 10 characters)'
+                helpText: ''
```
BrunoBernardino: Does this pass `zapier validate` ? I was initially thinking we didn't add the property at all, but if this still passes validation, I'm ok with it.

eliangcs: Nice catch! It does not pass `zapier validate`. Fixed in 01995e7.

PR #182, opened 2017-11-14
zapier convert: Escape special chars in field help text

eliangcs:
* Fix redundant slashes when escaping line breaks
* Escape single quotes (')

### DIFF
--- a/src/tests/utils/convert.js
+++ b/src/tests/utils/convert.js
@@ -57,9 +57,19 @@ describe('convert render functions', () => {
       };
 
       const string = convert.renderField(wbDef, wbKey);
-      console.log(string);
       const field = s2js(string);
-      field.helpText.should.eql('line 1\\nline 2\\nline 3\\n');
+      field.helpText.should.eql('line 1\nline 2\nline 3\n');
+    });
+
+    it('should escape single quotes in help text', () => {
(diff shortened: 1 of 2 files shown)


src/tests/utils/convert.js
```diff
@@ -57,9 +57,19 @@ describe('convert render functions', () => {
       };
 
       const string = convert.renderField(wbDef, wbKey);
-      console.log(string);
       const field = s2js(string);
-      field.helpText.should.eql('line 1\\nline 2\\nline 3\\n');
+      field.helpText.should.eql('line 1\nline 2\nline 3\n');
```
BrunoBernardino: Doesn't this mean the converted helpText will keep the newlines (which will be a problem when we print them in the template)?

eliangcs: That's what I thought when I wrote `'line 1\\nline 2\\nline 3\\n'` originally in this test. But since we eval `string` with `s2js` to an object, the escaped newlines in the template are unescaped again.

For example:
```
> eval("({helpText: 'line 1\\nline 2'})");
{ helpText: 'line 1\nline 2' }
```

Try converting an app that uses multi-lines help text and you'll see how this PR fixes the issue.

PR #183, opened 2017-11-14
zapier convert: Normalize newlines in scripting

eliangcs:
This PR normalizes all the newline characters to `\n` if user uses `\r\n` or `\r` as newlines in their JavaScript scripting.


BrunoBernardino:
BTW, If we could add a small little test for this, it'd be great.

eliangcs:
Good idea! Added tests in e5848cf.

PR #192, opened 2017-11-24
Migrate user by email.

BrunoBernardino:
[Relevant Trello Card](https://trello.com/c/o9u4x7sL/187-migrate-user-by-email)

Cannot be merged until https://github.com/zapier/zapier/pull/14438 ships.

### DIFF
--- a/src/commands/migrate.js
+++ b/src/commands/migrate.js
@@ -14,7 +14,11 @@ const migrate = (context, fromVersion, toVersion, optionalPercent = '100%') => {
       const user = global.argOpts.user;
 
       if (user) {
-        body.percent = 100;
+        if (optionalPercent !== 100) {
+          throw new Error('Cannot define percent and user. Use only one or the other.');
+        }
+
+        body.percent = optionalPercent;
         body.user = user;
 
         context.line(`Getting ready to migrate "${user}" in your app "${app.title}" from ${fromVersion} to ${toVersion}.\n`);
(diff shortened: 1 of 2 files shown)


src/commands/migrate.js
```diff
@@ -8,26 +8,43 @@ const migrate = (context, fromVersion, toVersion, optionalPercent = '100%') => {
...
-      context.line(`Getting ready to migrate your app "${app.title}" from ${fromVersion} to ${toVersion}.\n`);
-      utils.printStarting(`Starting migration from ${fromVersion} to ${toVersion} for ${optionalPercent}%`);
+      const body = {
+        percent: optionalPercent,
+      };
+      const user = global.argOpts.user;
+
+      if (user) {
+        body.percent = 100;
+        body.user = user;
```
eliangcs: Command-line parameters `percent` and `user` shouldn't co-exist. So I think we should block users if they specify both the parameters instead of setting `percent` to 100 silently. What do you think?

BrunoBernardino: I can see how that makes sense (maybe they did want the percentage instead). I'll make that change!

BrunoBernardino: Well, apparently we can't tell if the percentage was set or is the default one (since we default it to `100%`), so I don't think I can do what you're saying without changing the code around a bit, and it doesn't seem worth it.

What do you think?

eliangcs: How about we block users only if `percent != 100` and `user` is specified?

BrunoBernardino:
Thanks, I'll merge tomorrow, because https://github.com/zapier/zapier/pull/14438 was merged today but hasn't shipped yet.

PR #193, opened 2017-11-27
fix "should is not a function" bug on test run

mariyadiminsky:
When creating an app with the cli via `zapier init example-app`, a test file is created, ie. `test/index.js`. If a user testing out the cli doesn't manually change `require('should');` to `const should = require('should')` it will cause an error when running `zapier test`, ie. `should is not a function`. Actually even in the docs(https://zapier.com/developer/documentation/v2/getting-started-cli/) it creates a `const` for this.

### DIFF
--- a/scaffold/convert/create-test.template.js
+++ b/scaffold/convert/create-test.template.js
@@ -1,4 +1,4 @@
-require('should');
+const should = require('should');
 
 const zapier = require('zapier-platform-core');
 
--- a/scaffold/convert/search-test.template.js
+++ b/scaffold/convert/search-test.template.js
@@ -1,4 +1,4 @@
-const = require('should');
+const should = require('should');
 
 const zapier = require('zapier-platform-core');
 
--- a/scaffold/convert/search-test.template.js
+++ b/scaffold/convert/search-test.template.js
@@ -1,4 +1,4 @@
-require('should');
+const = require('should');
 
 const zapier = require('zapier-platform-core');
 
--- a/scaffold/convert/trigger-test.template.js
+++ b/scaffold/convert/trigger-test.template.js
@@ -1,4 +1,4 @@
-const = require('should');
+const should = require('should');
 
 const zapier = require('zapier-platform-core');
 
--- a/scaffold/convert/trigger-test.template.js
+++ b/scaffold/convert/trigger-test.template.js
@@ -1,4 +1,4 @@
-require('should');
+const = require('should');
 
 const zapier = require('zapier-platform-core');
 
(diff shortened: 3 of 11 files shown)


xavdid:
Yeah, `should` is a little funky like that. When required, it modifies the global object (typically an anti-pattern). That's why it's sometimes included in files that don't use the function, but do use `thing.should.be.true()`. It's only sometimes needed as a const. 

pertinent docs: 
> By default (when you `require('should')`) `should` extends the `Object.prototype` with a single non-enumerable getter that allows you to express how that object should behave. It also returns itself when required with require.

That said, I don't *think* we can pull it out of files. Tests are passing because as long as something requires `should`, other uses of it work. For any file to work in isolation, it has to repeat that work.

FWIW, I couldn't get an error in an example project. The following works fine: 

```bash
% zapier init example
% cd example
% npm i
% zapier test
```

Were you doing different steps?

mariyadiminsky:
@xavdid Hmm, I see.

Per the steps, I'm doing the exact same steps. This is the test file:
```
require('should');

const zapier = require('zapier-platform-core');

// Use this to make test calls into your app:
const App = require('../index');
const appTester = zapier.createAppTester(App);

describe('My App', () => {

  it('should load recipes', (done) => {
    const triggerPointer = 'triggers.recipe';
    const bundle = {
      inputData: {
        style: 'mediterranean'
      }
    };

    appTester(App.triggers.recipe.operation.perform, bundle)
      .then((results) => {
        should(results.length).above(1);

        const firstResult = results[0];
        console.log('test result: ', firstResult)
        should(firstResult.name).eql('name 1');
        should(firstResult.directions).eql('directions 1');

        done();
      })
      .catch(done)
  });

});
```

When you run `zapier test` you don't get an error?

xavdid:
Ah, mine is different. I got

```js
// example % cat test/index.js
require('should');

const zapier = require('zapier-platform-core');

// Use this to make test calls into your app:
const App = require('../index');
const appTester = zapier.createAppTester(App);

describe('My App', () => {

  it('should test something', (done) => {
    const x = 1;
    x.should.eql(1);
    done();
  });

});
```

Are you using a template or the plain `init`?

Either way, it's not good that our default (whichever template it is) doesn't work out of the box. I'm just not sure you call pull out the other requires.

mariyadiminsky:
@xavdid Ok so I'm going to close this PR. The reason this was happening was because in my example, following the directions here(https://zapier.com/developer/documentation/v2/getting-started-cli/) I was calling `should` as a function, and in your example it passes for me to because it's being called on `x` as a method. Maybe somehow the way `should` is implemented it won't work if called by itself, rather it needs to be called on an item.

PR #194, opened 2017-11-28
Allow reading from a changelog for app promotion.

BrunoBernardino:
Also adds tests.

[Relevant Trello Card](https://trello.com/c/vkFrD99V/148-require-allow-cli-developers-to-enter-a-change-message-when-promoting-a-version-to-production)

Cannot be merged until this ships, to support it in the backend: https://github.com/zapier/zapier/pull/14480

No changelog:
screen shot 2017-11-28 at 15 43 59

With changelog:
screen shot 2017-11-28 at 15 44 59

(now it asks for confirmation before continuing with or without the changelog)

### DIFF
--- a/src/commands/promote.js
+++ b/src/commands/promote.js
@@ -10,15 +10,29 @@ const promote = (context, version) => {
   let appId = 0;
 
   return utils.checkCredentials()
-    .then(() => utils.getLinkedApp())
-    .then((app) => {
+    .then(() => Promise.all([
+      utils.getLinkedApp(),
+      utils.getVersionChangelog(version),
+    ]))
+    .then(([app, changelog]) => {
       context.line(`Preparing to promote version ${version} of your app "${app.title}".\n`);
+
+      const body = {};
+
+      if (changelog) {
+        context.line(colors.green(`Changelog found for ${version}!`));
+        context.line(`\n---\n${changelog}\n---\n`);
+        body.changelog = changelog;
+      } else {
+        context.line(`${colors.yellow('Warning!')} Changelog not found. Please create a \`CHANGELOG.md\` file in a format similar to ${colors.cyan('https://github.com/zapier/zapier-platform-cli/blob/master/CHANGELOG.md')}, with user-facing descriptions.\n`);
+      }
+
       const url = `/apps/${app.id}/versions/${version}/promote/production`;
       appId = app.id;
       utils.printStarting(`Promoting ${version}`);
       return utils.callAPI(url, {
         method: 'PUT',
-        body: {}
+        body,
       }, false);
     })
     .then(() => {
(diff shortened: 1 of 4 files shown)


src/commands/promote.js
```diff
@@ -10,15 +10,29 @@ const promote = (context, version) => {
...
+      const body = {};
+
+      if (changelog) {
+        context.line(colors.green(`Changelog found for ${version}!`));
+        context.line(`\n---\n${changelog}\n---\n`);
+        body.changelog = changelog;
+      } else {
+        context.line(`${colors.yellow('Warning!')} Changelog not found. Please create a \`CHANGELOG.md\` file in a format similar to ${colors.cyan('https://github.com/zapier/zapier-platform-cli/blob/master/CHANGELOG.md')}, with user-facing descriptions.\n`);
+      }
+
```
eliangcs: Does it mean when a changelog is absent, we still go ahead and promote the app right away? I think we should ask the user here if they want to continue without a changelog. And even if a changelog is found, we should also print the changelog and ask the user if they want to continue to give them a chance to edit the changelog.

src/tests/utils/changelog.js
```diff
@@ -0,0 +1,31 @@
...
+    it('should not fail when there is no changelog file', () =>
+      changelog.getVersionChangelog('1.0.0', './unexistent-directory')
+        .then((log) => {
+          log.should.eql('');
+        })
+    );
+
+  });
+
+});
```
eliangcs: Love the tests here!

src/commands/promote.js
BrunoBernardino: That's a great point! For now we don't have that concept of an interactive command so I didn't want to create that just for this, especially because eventually we'll make changelog required (in some future major version), and we'll start throwing an error when it's not found.

I hope that makes sense!

eliangcs: Maybe I miss a context here. But don't we already have `zapier link` command, which asks the user which app to link?

BrunoBernardino: Oh 🤦‍♂️, you're right. I'll look into that, then!

BrunoBernardino: Alright, added!

src/commands/promote.js
```diff
@@ -1,24 +1,65 @@
...
+    .then(([foundApp, foundChangelog]) => {
+      app = foundApp;
+      changelog = foundChangelog;
+
       context.line(`Preparing to promote version ${version} of your app "${app.title}".\n`);
+
+      if (changelog) {
+        context.line(colors.green(`Changelog found for ${version}!`));
+        context.line(`\n---\n${changelog}\n---\n`);
+        return Promise.resolve('y');
```
eliangcs: Actually, I was thinking even if we found a changelog, the user may want to preview the changelog before they submit it for promotion. For example, a user may forget to put a blank line between `## 1.0.0` and its first changelog item. They may want to go back and fix it. So I think it's nice to ask the user if they want to proceed to promote with the changelog we parse out from `CHANGELOG.md`. What do you think?

BrunoBernardino: Oh, that's a good idea. I'll add that tomorrow then!

BrunoBernardino:
Can be merged when https://github.com/zapier/zapier/pull/14502 goes out.

PR #196, opened 2017-12-05
zapier convert: Add custom input field support

eliangcs:
This PR adds [custom input field](https://zapier.github.io/zapier-platform-cli/#customdynamic-fields) support to `zapier convert`, specifically for searches and creates.

### DIFF
--- a/scaffold/convert/search.template.js
+++ b/scaffold/convert/search.template.js
@@ -786,6 +786,98 @@ const getList = (z, bundle) => {
     });
 <% } %>
 };
+<% }
+
+if (inputFieldFullScripting) { %>
+const getInputFields = (z, bundle) => {
+  const scripting = require('../scripting');
+  const legacyScriptingRunner = require('zapier-platform-legacy-scripting-runner')(scripting);
+
+  bundle._legacyUrl = '<%= CUSTOM_FIELDS_URL %>';
+  bundle._legacyUrl = replaceVars(bundle._legacyUrl, bundle, {});
+
+  // Do a _custom_action_fields() from scripting.
+  const fullFieldsEvent = {
+    name: 'create.input',
+    key: '<%= KEY %>'
+  };
+  return legacyScriptingRunner.runEvent(fullFieldsEvent, z, bundle);
+};
+<% } else if (inputFieldPreScripting && !inputFieldPostScripting) { %>
+const getInputFields = (z, bundle) => {
+  const scripting = require('../scripting');
+  const legacyScriptingRunner = require('zapier-platform-legacy-scripting-runner')(scripting);
+
+  bundle._legacyUrl = '<%= CUSTOM_FIELDS_URL %>';
+  bundle._legacyUrl = replaceVars(bundle._legacyUrl, bundle, {});
+
+  // Do a _pre_custom_action_fields() from scripting.
+  const preFieldsEvent = {
+    name: 'create.input.pre',
+    key: '<%= KEY %>'
+  };
+  return legacyScriptingRunner.runEvent(preFieldsEvent, z, bundle)
+    .then((preFieldsResult) => z.request(preFieldsResult))
+    .then((response) => z.JSON.parse(response.content));
+};
+<% } else if (inputFieldPreScripting && inputFieldPostScripting) { %>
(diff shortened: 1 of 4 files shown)


src/utils/convert.js
```diff
@@ -338,13 +377,34 @@ const getStepMetaData = (definition, type, key) => {
...
+  const hasResourcePostScripting = hasScriptingMethod(js, type, key, 'post', 'resource');
+  const hasResourceFullScripting = hasScriptingMethod(js, type, key, 'full', 'resource');
+
+  const hasInputFieldPreScripting = hasScriptingMethod(js, type, key, 'pre', 'input_fields');
+  const hasInputFieldPostScripting = hasScriptingMethod(js, type, key, 'post', 'input_fields');
+  const hasInputFieldFullScripting = hasScriptingMethod(js, type, key, 'full', 'input_fields');
+
+  const hasOutputFieldPreScripting = hasScriptingMethod(js, type, key, 'pre', 'output_fields');
+  const hasOutputFieldPostScripting = hasScriptingMethod(js, type, key, 'post', 'output_fields');
+  const hasOutputFieldFullScripting = hasScriptingMethod(js, type, key, 'full', 'output_fields');
```
eliangcs: I've refactored here a bit so that we use `hasScriptingMethod()` to get all `hasXXXScripting` booleans. That means no direct calls to `js.indexOf(methodName)`.

src/utils/convert.js
```diff
@@ -291,43 +298,75 @@ const writeAuth = (definition, newAppDir) => {
...
+  const methodSuffix = suffixTable[type][position][method_type];
+  if (!methodSuffix) {
+    return false;
   }
 
+  // Whole word search. '\b' regex matches a word boundary.
   const methodName = `${key}${methodSuffix}`;
-
-  return (js.indexOf(methodName) !== -1);
+  return new RegExp('\\b' + methodName + '\\b').test(js);
```
eliangcs: Using `js.indexOf(methodName)` to determine a method exists can be wrong if we have a method named `foo_bar` and want to see if `foo` exists. Here I've fixed it with regex whole word searching.

src/utils/convert.js
```diff
@@ -291,43 +298,75 @@ const writeAuth = (definition, newAppDir) => {
...
+        output_fields: '_post_custom_search_result_fields',
+      },
+      full: {
+        step: '_search',
+        resource: '_read_resource',
+        input_fields: '_custom_search_fields',
+        output_fields: '_custom_search_result_fields',
+      },
+    },
+  };
```
eliangcs: Refactored the spaghetti code with a mapping table.

scaffold/convert/search.template.js
```diff
@@ -786,6 +786,98 @@ const getList = (z, bundle) => {
...
+
+if (inputFieldFullScripting) { %>
+const getInputFields = (z, bundle) => {
+  const scripting = require('../scripting');
+  const legacyScriptingRunner = require('zapier-platform-legacy-scripting-runner')(scripting);
+
+  bundle._legacyUrl = '<%= CUSTOM_FIELDS_URL %>';
+  bundle._legacyUrl = replaceVars(bundle._legacyUrl, bundle, {});
+
+  // Do a _custom_action_fields() from scripting.
```
BrunoBernardino: This comment should mention `_custom_search_fields()`, and I noticed the same pattern below.

scripts/test-convert.js
```diff
@@ -14,6 +14,7 @@ const appsToConvert = [
   {id: 82052, name: 'simple-oauth'},
   {id: 82250, name: 'search-oauth'},
   {id: 82251, name: 'basic-api-header'},
+  {id: 82460, name: 'custom-fields'},
```
BrunoBernardino: Nice!

src/utils/convert.js
BrunoBernardino: Much simpler and nice! Good work!

src/utils/convert.js
BrunoBernardino: Nice improvement!

eliangcs:
Nice catch. I forgot to change them after a copy-and-paste. Fixed in a30c38f.

PR #198, opened 2017-12-06
Log the user in if they try to push before they're logged in

### LINKED ISSUE #119: zapier push does not check .zapierrc until after dev names an app
I ran `zapier push` without logging in. I was prompted to name my app and *then* told to login.

A better first experience would be to walked through the login process OR told to login before I create the name for my app.

xavdid:
Fixes #119 

The only weird part is I tweaked `login` to take a default param, since the line about starting a new app with `zapier init .` doesn't always make sense. It's there by default, but we can log users in without it surfacing.


xavdid:
Great, thanks!

PR #199, opened 2017-12-07
zapier convert: Add custom output field support

eliangcs:


scaffold/convert/create.template.js
```diff
@@ -202,6 +202,98 @@ const getInputFields = (z, bundle) => {
...
 };
+<% }
+
+if (outputFieldFullScripting) {%>
+const getOutputFields = (z, bundle) => {
+  const scripting = require('../scripting');
+  const legacyScriptingRunner = require('zapier-platform-legacy-scripting-runner')(scripting);
+
+  bundle._legacyUrl = '<%= CUSTOM_FIELDS_RESULT_URL %>';
+  bundle._legacyUrl = legacyScriptingRunner.replaceVars(bundle._legacyUrl, bundle, {});
```
BrunoBernardino: Since we're creating a "local" `replaceVars`, I think it makes sense to use that here (and making it consistent with input fields).

This happens in a few places below as well.

src/utils/convert.js
```diff
@@ -400,11 +400,13 @@ const getStepMetaData = (definition, type, key) => {
 
   const hasCustomInputFields = (
     hasInputFieldPreScripting || hasInputFieldPostScripting || hasInputFieldFullScripting ||
-    Boolean(stepDef.custom_fields_url)
+    (type !== 'trigger' && Boolean(stepDef.custom_fields_url))
+    // Triggers don't have custom input fields
```
BrunoBernardino: I'd prefer to be clearer and avoid confusion here, by changing this comment to something like: `// Triggers in WB don't have custom input fields`

eliangcs: Agreed! Fixed in 746f9de.

scaffold/convert/create.template.js
eliangcs: Fixed in 7a5d22f. And since we're using `replaceVars()` in several places, I think it's a good idea to separate the function to a shared utility module.

PR #200, opened 2017-12-08
zapier convert: Add support for basic auth mapping

eliangcs:
This PR handles the case where a WB app with basic authentication has an [authentication mapping](https://zapier.com/developer/documentation/v2/authentication-mappings/). In CLI, we don't have authentication mappings. Instead, we just let users freely define their authentication in a function. So when converting a WB app like this, we need to use  `custom` auth type and mimic the mapping of the auth fields in the authentication function.

Example app to try out: [80444](https://zapier.com/developer/builder/app/80444/)

### DIFF
--- a/scaffold/convert/header.template.js
+++ b/scaffold/convert/header.template.js
@@ -1,5 +1,6 @@
+<% if (customBasic) { %>
 const { replaceVars } = require('./utils');
-
+<% } %>
 <% if (before && !session && !oauth && !customBasic) { %>const maybeIncludeAuth = (request, z, bundle) => {
 <%
   Object.keys(mapping).forEach((mapperKey) => {
--- a/scaffold/convert/header.template.js
+++ b/scaffold/convert/header.template.js
@@ -1,4 +1,6 @@
-<% if (before && !session && !oauth) { %>const maybeIncludeAuth = (request, z, bundle) => {
+const { replaceVars } = require('./utils');
+
+<% if (before && !session && !oauth && !customBasic) { %>const maybeIncludeAuth = (request, z, bundle) => {
 <%
   Object.keys(mapping).forEach((mapperKey) => {
     fields.forEach((field) => {
@@ -12,6 +14,18 @@
     });
   });
 %>
+  return request;
+};
+<% } else if (customBasic) { %>
+const maybeIncludeAuth = (request, z, bundle) => {
+  const mapping = {
+    username: '<%= mapping.username %>',
+    password: '<%= mapping.password %>'
+  };
+  const username = replaceVars(mapping.username, bundle);
+  const password = replaceVars(mapping.password, bundle);
+  const encoded = `${username}:${password}`.toString('base64');
+  request.headers.Authorization = `Baisc ${encoded}`;
   return request;
 };
 <% }
--- a/scaffold/convert/header.template.js
+++ b/scaffold/convert/header.template.js
@@ -25,7 +25,7 @@ const maybeIncludeAuth = (request, z, bundle) => {
   };
   const username = replaceVars(mapping.username, bundle);
   const password = replaceVars(mapping.password, bundle);
-  const encoded = `${username}:${password}`.toString('base64');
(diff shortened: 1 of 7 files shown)


scaffold/convert/header.template.js
```diff
@@ -1,4 +1,7 @@
-<% if (before && !session && !oauth) { %>const maybeIncludeAuth = (request, z, bundle) => {
+<% if (customBasic) { %>
```
eliangcs: I use the term "custom basic" to indicate a basic authentication that does not use username and password as user-facing authentication fields.

src/tests/utils/convert.js
```diff
@@ -120,7 +120,7 @@ describe('convert render functions', () => {
       const wbDef = definitions.noAuth;
       return convert.renderIndex(wbDef)
         .then(string => {
-          string.should.containEql('authentication: {}');
+          string.should.not.containEql('authentication:');
```
eliangcs: Originally, we generated an empty `{}` when a WB app doesn't have auth. The right way is not to include an `authentication` section at all if the WB app doesn't have auth.

src/tests/utils/convert.js
```diff
@@ -202,14 +202,12 @@ describe('convert render functions', () => {
 
       return convert.getHeader(wbDef)
         .then(string => {
-          string.should.eql(`const maybeIncludeAuth = (request, z, bundle) => {
+          string.trim().should.eql(`const maybeIncludeAuth = (request, z, bundle) => {
```
eliangcs: `.trim()` is to remove leading and trailing spaces generated by the template. I leave the code unformatted because, instead of formatting the code in the templates, I think I'll make another PR that uses [Prettier](https://prettier.io) to format the generated code.

src/utils/convert.js
```diff
@@ -95,7 +95,7 @@ const getAuthType = (definition) => {
 };
 
 const hasAuth = (definition) => {
-  return getAuthType(definition) !== 'custom' || !_.isEmpty(definition.auth_fields);
+  return getAuthType(definition) !== 'custom' && !_.isEmpty(definition.auth_fields);
```
eliangcs: Fixes the logic determining if an app has auth.

src/utils/convert.js
```diff
@@ -585,15 +598,20 @@ const writeStep = (type, definition, key, legacyApp, newAppDir) => {
...
+      const upperKey = key.toUpperCase();
+      return `        ${key}: process.env.${upperKey}`;
+    });
     result = `{
-        username: process.env.USERNAME,
-        password: process.env.PASSWORD
+${lines.join(',\n')}
       }`;
     break;
+  }
```
eliangcs: In the generated test code, we were assuming basic auth is always username and password, which is not true. We should really take advantage of `auth_fields` instead.

bcooksey:
Will try to test this out a bit tomorrow. Off the cuff, does this handle a case where say an app only has an `api_key` for auth fields and is doing something like `Authorization: Basic <api_key>:`?

eliangcs:
@bcooksey exactly that case!

scaffold/convert/header.template.js
```diff
@@ -14,6 +17,18 @@
...
 };
+<% } else if (customBasic) { %>
+const maybeIncludeAuth = (request, z, bundle) => {
+  const mapping = {
+    username: '<%= mapping.username %>',
+    password: '<%= mapping.password %>'
+  };
+  const username = replaceVars(mapping.username, bundle);
+  const password = replaceVars(mapping.password, bundle);
+  const encoded = `${username}:${password}`.toString('base64');
```
bcooksey: `toString()` on a string won't encode in JS like you'd hope. Need to wrapper in a Buffer I believe.

```
> 'asdf:'.toString('base64')
'asdf:'
> new Buffer('asdf:').toString('base64')
'YXNkZjo='
```

eliangcs: @bcooksey nice catch! Fixed in 7ac3941.

PR #204, opened 2017-12-12
zapier convert: Exclude input fields from bundle.inputData if send_in_json is false

eliangcs:
This PR adds support for selectively excluding input fields from `bundle.inputData` in creates that don't have `_write` nor `_pre_write` scripting, where the fields have `send_in_json` set to false. In WB, these fields weren't included, and in the current conversion they are.

Example app to try out: [83073](https://zapier.com/developer/builder/app/83073/development)

BTW, this PR also fixes the crash issue when a field label contains a single quote.


eliangcs:
@xavdid thanks for the reminder! I just double checked. The checkbox is not available for triggers and searches.

PR #207, opened 2017-12-14
zapier convert: Add support for search-or-creates

eliangcs:
This PR adds an entry to `App.searchOrCreates` whenever it finds a search paired with a create.

Example app to try out: [83195](https://zapier.com/developer/builder/app/83195/development)

### DIFF
--- a/scripts/test-convert.js
+++ b/scripts/test-convert.js
@@ -15,6 +15,7 @@ const appsToConvert = [
   {id: 82250, name: 'search-oauth'},
   {id: 82251, name: 'basic-api-header'},
   {id: 82460, name: 'custom-fields'},
+  {id: 83195, name: 'search-or-create'},
   // TODO: Add more apps that use different scripting methods, as we start to support them
 ];
 
--- a/scripts/test-convert.js
+++ b/scripts/test-convert.js
@@ -16,7 +16,7 @@ const appsToConvert = [
   { id: 82251, name: 'basic-api-header' },
   { id: 82460, name: 'custom-fields' },
   { id: 83195, name: 'search-or-create' },
-  { id: 80444, name: 'custom-baisc' },
+  { id: 80444, name: 'custom-basic' },
   { id: 83073, name: 'send-in-json' },
   { id: 83342, name: 'replace-vars' }
   // TODO: Add more apps that use different scripting methods, as we start to support them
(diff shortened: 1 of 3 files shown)


scripts/test-convert.js
```diff
@@ -15,8 +15,9 @@ const appsToConvert = [
   { id: 82250, name: 'search-oauth' },
   { id: 82251, name: 'basic-api-header' },
   { id: 82460, name: 'custom-fields' },
+  { id: 83195, name: 'search-or-create' },
+  { id: 80444, name: 'custom-baisc' },
```
xavdid: `basic` typo, possibly related to a bad merge?

src/utils/convert.js
```diff
@@ -876,6 +899,13 @@ const renderIndex = legacyApp => {
 
     templateContext.REQUIRES = importLines.join('\n');
 
+    const searchOrCreates = findSearchOrCreates(legacyApp);
+    if (_.isEmpty(searchOrCreates)) {
+      templateContext.SEARCH_OR_CREATES = null;
```
xavdid: do we need to set this explicitly? It'll be `undefined` by default until it's set. Could just have `not _.isEmpty` as the only branch.

src/utils/convert.js
```diff
@@ -876,6 +899,13 @@ const renderIndex = legacyApp => {
...
+    } else {
+      templateContext.SEARCH_OR_CREATES = JSON.stringify(searchOrCreates, null, 2);
```
xavdid: super minor, but maybe worth moving the `2` a `spacing` variable somewhere so we can change it all in one go if we ever need to.

scripts/test-convert.js
eliangcs: Yeah, it came from the merge. Fixed in 6909cc4. Thanks!

src/utils/convert.js
eliangcs: Unfortunely, we can't do that because `_.template()` throws an exception when a variable is undfined in the template.

src/utils/convert.js
eliangcs: I don't follow you on this one. Could you elaborate?

src/utils/convert.js
xavdid: ah, got it! Thanks for the explanation

src/utils/convert.js
xavdid: I meant something like

```js
const JSON_SPACING = 2;
// ...
JSON.stringify(searchOrCreates, null, JSON_SPACING);
```

but it turns out we're only calling `stringify` twice in the file, so it's not a big deal.

PR #208, opened 2017-12-15
zapier convert: Replace variables in URLs (PDE-10)

eliangcs:
Developers can put variables from in a URL, such as `https://{{account}}.example.com/{{resource_name}}`. This PR makes sure we replace those variables in URLs before we send requests.

Example app to try out: [83342](https://zapier.com/developer/builder/app/83342/development):
```
zapier convert 83342 replace-vars
cd replace-vars
npm install
zapier test
```

Expect `zapier test` to fail on a TypeError for test case `Full Search` and `Full Trigger` due to a bug in core, which was addressed in zapier/zapier-platform-core#59.

### DIFF
--- a/scaffold/convert/create-test.template.js
+++ b/scaffold/convert/create-test.template.js
@@ -11,9 +11,9 @@ describe('Creates - <%= LABEL %>', () => {
   it('should create an object', (done) => {
     const bundle = {
       authData: <%= AUTH_DATA %>,
-      inputData: {
-        name: 'Testing',
-      },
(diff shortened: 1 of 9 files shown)


ZaneLyon:
I'm testing the Jira integration, so I added a Jira task reference to the title of this PR to see if linking is working properly.

eliangcs:
@xavdid yeah, you guess it right. You get the same result whether you pass a `{}` or an`undefined` to `replaceVars`, so I removed it to make the code cleaner.

PR #188, opened 2017-11-20
Customize update notification message

BrunoBernardino:
Basically add a reminder to re-test the integration after an update.

[Trello Card](https://trello.com/c/KDQ7C7Q7/212-improve-update-notification-on-cli).

screen shot 2017-11-20 at 16 32 37

PR #210, opened 2017-12-18
use default update interval (daily)

xavdid:
See [this comment](https://github.com/zapier/zapier-platform-cli/pull/188#issuecomment-352499434). We don't want to bug users every single time they run a CLI command.

### DIFF
--- a/src/entry.js
+++ b/src/entry.js
@@ -27,10 +27,7 @@ module.exports = argv => {
     process.exit(1);
   }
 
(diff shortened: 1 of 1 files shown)


xavdid:
Yeah, I'm not sure why the build command is taking so long recently. Did we change something there?

eliangcs:
@xavdid not that I'm aware of. We didn't change the build command recently.

PR #212, opened 2017-12-20
scaffold also creates a test

xavdid:
see: https://trello.com/c/8h7hvYV2

As a side note, it would be nice if our scaffolded material used a url that actually existed, so tests would run. in my testing for this, i used `https://jsonplaceholder.typicode.com/`, which accepts all manner of verbs and params.

### DIFF
--- a/scaffold/test.template.js
+++ b/scaffold/test.template.js
@@ -1,0 +1,22 @@
+/* globals describe it */
+const should = require('should');
+
+const zapier = require('zapier-platform-core');
+
+// Use this to make test calls into your app:
+const App = require('../../index');
+const appTester = zapier.createAppTester(App);
+zapier.tools.env.inject();
+
+describe('My App', () => {
+  it('should run <%= TYPE %>s.<%= KEY %>', done => {
+    const bundle = { inputData: {} };
+
+    appTester(App.<%= TYPE %>s.<%= KEY %>.<%= MAYBE_RESOURCE %>operation.perform, bundle)
+      .then(results => {
+        should.exist(results)
+        done();
+      })
+      .catch(done);
+  });
+});
--- a/scaffold/test.template.js
+++ b/scaffold/test.template.js
@@ -12,9 +12,9 @@ describe('My App', () => {
   it('should run <%= TYPE %>s.<%= KEY %>', done => {
     const bundle = { inputData: {} };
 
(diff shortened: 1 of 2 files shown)


scaffold/test.template.js
```diff
@@ -0,0 +1,22 @@
...
+const appTester = zapier.createAppTester(App);
+zapier.tools.env.inject();
+
+describe('My App', () => {
+  it('should run <%= TYPE %>s.<%= KEY %>', done => {
+    const bundle = { inputData: {} };
+
+    appTester(App.<%= TYPE %>s.<%= KEY %>.<%= MAYBE_RESOURCE %>operation.perform, bundle)
+      .then(results => {
+        should.exist(results)
```
eliangcs: Missed a semicolon at the end of line.

scaffold/test.template.js
```diff
@@ -0,0 +1,22 @@
...
+// Use this to make test calls into your app:
+const App = require('../../index');
+const appTester = zapier.createAppTester(App);
+zapier.tools.env.inject();
+
+describe('My App', () => {
+  it('should run <%= TYPE %>s.<%= KEY %>', done => {
+    const bundle = { inputData: {} };
+
+    appTester(App.<%= TYPE %>s.<%= KEY %>.<%= MAYBE_RESOURCE %>operation.perform, bundle)
```
eliangcs: If `TYPE` is `search`, this would render `App.searchs`, which should be `App.searches`.

scaffold/test.template.js
xavdid: good eye! Since these don't parse as valid JS, i'm flying a little blind without my tools, haha.

xavdid:
I looked through the scaffold tests and I think we'd want to go a little different direction with these. I think we can make the scaffolds really specific (maybe to the point of including checks for params or something) while the scaffolded tests should be as generic as possible (hence there being just the one test for all step types). So great suggestion, but I think we should keep two versions

reganstarr:
@xavdid I just had a "wow moment" when I saw that tests were auto-created when I ran `scaffold`. This feature was on my wish-list and I had to track down and thank whoever made it happen. So thank you! :)

xavdid:
haha no problem @reganstarr! Glad it's getting used. :)

PR #213, opened 2017-12-20
zapier convert: Use Buffer.from() instead of Buffer constructor

eliangcs:
See https://eslint.org/docs/rules/no-buffer-constructor


eliangcs:
@xavdid I found another one and fixed it in 78ac665. Feel free to grab the others in core. Thanks!

PR #214, opened 2017-12-20
Increase test time limit

eliangcs:
Attempts to fix the following error on Travis:

```
1) build should list only required files:

  Error: Timeout of 15000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```


eliangcs:
@xavdid agreed that this is only a short-term workaround. We should revisit that long-running test when we have time. Thanks!

PR #215, opened 2017-12-21
use actual json endpoint for scaffold triggers

xavdid:
Before this PR, the scaffolded steps all used `http://example.com/<key>.json` which looks nice, but returns busted html whenever it's called. Instead, I switched the scaffolded steps to use https://jsonplaceholder.typicode.com, which is a test utility that supports all manner of endpoints and verbs. I'm using the default public stuff right now, but it's open source and we could customize it and host it ourselves if we want. The biggest advantage to self-hosting would be we could create wildcard endpoints to match the scaffold name (`/whatever` would work), but this seems straightforward enough as not to confuse devs. 

Now, the scaffolded tests actually run!

Before: 
screen shot 2017-12-21 at 12 57 57 am

After:
screen shot 2017-12-21 at 12 58 16 am

### DIFF
--- a/scaffold/create.template.js
+++ b/scaffold/create.template.js
@@ -2,7 +2,7 @@
 const create<%= CAMEL %> = (z, bundle) => {
   const responsePromise = z.request({
     method: 'POST',
-    url: 'http://example.com/api/<%= KEY %>s.json',
+    url: 'https://jsonplaceholder.typicode.com/posts',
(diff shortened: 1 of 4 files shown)


xavdid:
good point! those are private right now, so if we're going to be sending these right to customers, I'd prefer something public. This is easy to tweak again if it's not working out though. Thanks for the 👀!

PR #224, opened 2018-01-10
Set file mode to 777 when zipping

FokkeZB:
Hopefully fixes https://github.com/zapier/zapier/issues/14929

I added a test for `makeZip` but couldn't find a way to test if permissions where applied. I can see it works when I extract via MacOS Finder, but when I extract via `adm-zip` or https://www.npmjs.com/package/extract-zip (which claims to use zip modes if found) the permissions are the defaults.


FokkeZB:
I've found an extractor that does preserve permissions and added tests to validate a file that has initially no group/public permissions does have them after unzipping.

package.json
```diff
@@ -13,18 +13,24 @@
   "scripts": {
     "docs": "npm run build && node lib/bin/docs.js",
     "preversion": "git pull && npm test",
-    "version": "npm run docs && npm run gen-completions && git add docs/* goodies/* README.md",
+    "version":
+      "npm run docs && npm run gen-completions && git add docs/* goodies/* README.md",
```
FokkeZB: Not sure if this is as attended, but on commit it started formatting these files.

xavdid: mostly intended. maybe should exclude this, since it gets reformatted when it's modified by npm

FokkeZB:
@bryanhelmig do I have your blessing?

FokkeZB:
I guess we could use `0o755` instead. This PR lets archiver apply the permissions. If we wanted to prompt the user then we'd need to check the permissions on each file ourselves.

PR #225, opened 2018-01-11
add examples section

xavdid:
Add some code examples for often-asked questions. XML is the big one, but multiple pages in polling came up recently too.

### DIFF
--- a/README-source.md
+++ b/README-source.md
@@ -1227,10 +1227,16 @@ A: Not natively, but it can! Users have reported that the following `npm` module
 
 *Q: Is it possible to iterate over pages in a polling trigger?*
 
-A: Yes, though there are caveats. Your entire function only gets 30 seconds to run. HTTP requests are costly, so paging through a list may timeout (which you should avoid at all costs). You should keep an the polling [guidelines](https://zapier.com/developer/documentation/v2/deduplication/), namely the part about only iterating until you hit items that weren't created in the last few hours.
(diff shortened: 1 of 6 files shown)


snippets/xml.js
```diff
@@ -0,0 +1,12 @@
+const xml = require('pixl-xml');
+
+const App = {
+  // ...
+  afterRequest: [
```
eliangcs: I think this should be `afterResponse`.

README-source.md
```diff
@@ -1211,7 +1211,29 @@ There are a lot of details left out - check out the full example app for a worki
...
+* [xml2js](https://github.com/Leonidas-from-XIV/node-xml2js)
+* [fast-xml-parser](https://github.com/NaturalIntelligence/fast-xml-parser)
+
+```javascript
+[insert-file:./snippets/xml.js]
+```
+
+*Q: Is it possible to iterate over pages in a polling trigger?*
+
+A: Yes, though there are caveats. Your entire function only gets 30 seconds to run. HTTP requests are costly, so paging through a list may timeout (which you should avoid at all costs). You should keep an the polling [guidelines](https://zapier.com/developer/documentation/v2/deduplication/), namely the part about only iterating until you hit items that weren't created in the last few hours.
```
eliangcs: > "You should keep **an** the polling guidelines..."

I believe we have a typo here, but I don't know what it is.

xavdid: Should be "an eye". Thanks, I'll fix tomorrow!

snippets/xml.js
xavdid: Yes it should!

PR #226, opened 2018-01-13
Escape line breaks when generating package.json

bradleypriest:
Running `zapier convert` for an existing Zapier app with a multiline description generated an invalid JSON in package.json which then causes the whole process to fail with something similar to 

```
Error! Unexpected token
 in JSON at position 149
```

### DIFF
--- a/src/utils/convert.js
+++ b/src/utils/convert.js
@@ -922,7 +922,7 @@ const renderPackageJson = legacyApp => {
 
   const templateContext = {
     NAME: _.kebabCase(legacyApp.general.title),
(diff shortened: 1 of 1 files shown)


xavdid:
Thanks for the contribution @bradleypriest! 

We'll need you to sign the CLA (@bcooksey is that bot around?) but once your changes are approved by the engineer who has been working on the converter, you should be good to go!

bcooksey:
Kicked it off. Should be able to click the details link in the checks section now @bradleypriest

PR #228, opened 2018-01-16
In Readme, update the endpoint port to localhost:8001

Mariatta:


Mariatta:
Thanks for the review :)

Issue #41, opened 2017-03-21
Validate should check that generated keys from resources do not collide with manually defined keys


bcooksey:
When Zapier turns a resource into corresponding triggers/searches/creates, we auto-compute a key for those generated operations. For example, if an app has a resource called  "contact", the key for that action/create would be "contactCreate". 

What happens if there's already an action defined with that key? We should have validate raise an error so the dev knows to give the standalone action a new key.

PR #229, opened 2018-01-18
WIP: first pass at key collision detection

xavdid:
Fixes [PDE-35](https://zapierorg.atlassian.net/browse/PDE-35) and #41. 

This runs pre-test and during `zapier validate`. Does **not** run during `zapier build`.

Need to add tests and clean up the UX - validation is getting pretty crowded.


xavdid:
jk, this change actually has to happen in core

PR #230, opened 2018-01-23
Allow never versions of Node.js

janaagaard75:
Allow never versions of Node.js instead of requiring the exact version 6.10.2. Making this change will make it possible to use Yarn to install packages.

Reproducing the error in Yarn:
```shell
# create a directory with the minimum required files:
zapier init example-app
# move into the new directory:
cd example-app
# Install packages using Yarn instead of npm:
yarn install
> error zapier-platform-example-app-minimal@1.0.0: The engine "node" is incompatible with this module. Expected version "6.10.2".
```

Changing `"node": "6.10.2",` to `"node": ">=6.10.2",` in `package.json` fixes this issue.

I must admit that I have not double checked that this code change does indeed add `>=` in front of the node version number in `package.json` when running `zapier init example-app`.


janaagaard75:
Just realized, that it's actually on purpose, that the version number has been fixed.

PR #231, opened 2018-01-25
doc cleaning and examples

### LINKED ISSUE #131: Way to paginate polling trigger
I'm having trouble enabling pagination of triggers that are polling based.

I was not able to find anything in the docs suggesting how is `bundle.meta.page` modified.

Related issue : https://github.com/zapier/zapier-platform-core/issues/42

I'm expecting to be able to create a trigger that will paginate until the first duplicate of id is encountered. 

So far I tried adding `canPaginate: true` on my trigger definition, but it seems that this switch doesn't do anything.

Thanks for your help.

### LINKED ISSUE #164: Clarify search-powered fields usage
We have developers confused over when they can/should use Search-Powered Fields and are trying to use them in Trigger steps. It would be good to clarify within the documentation that this is for use only with Action steps (and possibly refer them to dynamic drop-downs if that's what they're looking for).

https://zapier.github.io/zapier-platform-cli/#search-powered-fields

xavdid:
Some general changes and adding more examples. I'll run docs once it's 👍 to keep the PR a little cleaner. 

@bcooksey if you've got a sec, i'm a little foggy on how correct I am with what exactly the keys in `bundle.meta` do. I copied from [here](https://zapier.com/developer/documentation/v2/scripting/#extra-request-info-via-bundlemeta), but I remember some weirdness about the distinction between UI test vs auth test (or if that's relevant in CLI). Anecdotally, I had some test polls today that didn't set the flags I'd expect, so I'd like to set the record straight. Also, i'm super unclear on what `bundle.meta.limit` is and when it's relevant.

For [PDE-33](https://zapierorg.atlassian.net/browse/PDE-33). Fixes #131, fixes #164

### DIFF
--- a/README-source.md
+++ b/README-source.md
@@ -1209,13 +1209,14 @@ See [the wiki](https://github.com/zapier/zapier-platform-cli/wiki/Example-Apps) 
 
 ## FAQs
 
-*Q: Why doesn't Zapier support newer versions of Node.js?*
+### Why doesn't Zapier support newer versions of Node.js?
 
-A: We run your code on AWS Lambda, which only supports a few [versions](https://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html) of Node (the latest of which is `LAMBDA_VERSION`. As that updates, so too will we.
+We run your code on AWS Lambda, which only supports a few [versions](https://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html) of Node (the latest of which is `LAMBDA_VERSION`. As that updates, so too will we.
 
-*Q: Does Zapier support XML (SOAP) APIs?*
 
-A: Not natively, but it can! Users have reported that the following `npm` modules are compatible with the CLI Platform:
+### Does Zapier support XML (SOAP) APIs?
+
+Not natively, but it can! Users have reported that the following `npm` modules are compatible with the CLI Platform:
 
 * [pixl-xml](https://github.com/jhuckaby/pixl-xml)
 * [xml2js](https://github.com/Leonidas-from-XIV/node-xml2js)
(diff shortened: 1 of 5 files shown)


README-source.md
```diff
@@ -726,13 +711,13 @@ For example, you can access the `process.env` in your perform functions:
 There are two primary ways to make HTTP requests in the Zapier platform:
 
 1. **Shorthand HTTP Requests** - these are simple object literals that make it easy to define simple requests.
-1. **Manual HTTP Requests** - this is much less "magic", you use `z.request([url], options)` to make the requests and control the response.
+2. **Manual HTTP Requests** - you use `z.request([url], options)` to make the requests and control the response. Use this when you need to change options for certain requests (for all requests, use middleware)
```
eliangcs: Missed a period at the end.

README-source.md
```diff
@@ -587,42 +587,27 @@ This object holds the user's auth details and the data for the API requests.
...
-};
-```
+| key | default | description |
+| ------------- | ------- | ---------------------------------------- |
+| frontend | `false` | whether this poll was initiated via the zap editor |
+| prefill | `false` | whether this poll is being used as a dynamic dropdown |
+| hydrate | `true`  | whether the results of this poll will be hydrated |
+| test_poll | `false` | whether the poll was triggered by a user testing their account |
+| standard_poll| `true`  | the opposite of `test_poll` |
+| first_poll | `false` | whether the poll was an initial check for items when a zap is turned on (none of which will trigger) |
```
bcooksey: Whew, this is a tough one to summarize. Maybe we drop what's in the parens and put a link that explains deduplication? So that way when you read it, it will be "Whether the poll was an initial check for items when a Zap turns on. Read more about De-duplication."?

I think something needs to indicate that the items returned populate our de-dupe system.

README-source.md
```diff
@@ -587,42 +587,27 @@ This object holds the user's auth details and the data for the API requests.
...
+| limit | `-1`    | the number of items to fetch for performance reasons. `-1` indicates there's no limit |
```
bcooksey: Can probably drop the "for performance reasons"

xavdid: do we use this for anything? It truncates the results that come in, but I don't understand why we'd use that

https://github.com/zapier/zapier/blob/develop/api/resources_flow.py#L1390-L1391

README-source.md
xavdid: good call, added a dedup section.

xavdid:
cool, thanks for the feedback! I added some extra sections and it all makes a lot more sense now.

README-source.md
bcooksey: It's going to be more important when Better Sample Data launches. They'll be able to use it to say "grab 3 records" for samples in the editor, vs a -1 which might cause us to grab 100 records and throw 97 away

PR #234, opened 2018-02-09
No version lock

xavdid:
Addressed [PDE-48](https://zapierorg.atlassian.net/browse/PDE-48) by not requiring a specific node version to run tests or build. Also introduces linting for unsupported features, which seems to work reasonably well.

This removes the version checks, adds linting, cleans up the `.npmignore`, and updates docs. I'm not sure what I want to do with 
the `.nvmrc` files (whether its worth generating them during `zapier init` or not), so if anyone has thoughts, I'd love to hear them. I don't *think* they matter as much anymore, so might be worth removing. 

If you want to test manually, I've been using the following trigger, which when included, won't build: 

```js
module.exports = {
  key: 'exponent',
  noun: 'Exponent',

  display: {
    label: 'Get Exponent',
    description: 'Triggers on a new exponent.'
  },

  operation: {
    inputFields: [
      { key: 'base', type: 'number', required: true },
      { key: 'exponent', type: 'number' }
    ],
    perform: (z, bundle) => {
      const letters = { a: 1, b: 2, c: 3 }
      return [
        {
          id: 1,
          res: bundle.inputData.base ** (bundle.inputData.exponent || 2),
          letters: Object.values(letters)
        }
      ]
    }
  }
}
```


xavdid:
Oh and as a side note, we probably don't need the nvmrc in this repo to be 6 no matter what we do with this PR. since we're using babel, might as well use async/await and some other good stuff.

src/utils/build.js
```diff
@@ -106,6 +108,39 @@ const listFiles = dir => {
...
+
+  // only lint js files; node_modules ignored by default
+  const jsPaths = paths.filter(p => p.endsWith('.js'));
+  const cli = new eslint.CLIEngine(opts);
+  const errors = eslint.CLIEngine.getErrorResults(
+    cli.executeOnFiles(jsPaths).results
+  );
+
+  if (errors.length) {
+    if (!process.env.NODE_ENV === 'test') {
```
eliangcs: I think you meant to write `process.env.NODE_ENV !== 'test'`. Otherwise, the errors will never be printed.

README-source.md
```diff
@@ -58,16 +58,12 @@ All Zapier CLI apps are run using Node.js `LAMBDA_VERSION`.
 
 You can develop using any version of Node you'd like, but your code has to run on Node `LAMBDA_VERSION`. You can accomplish this by developing on your preferred version and then transpiling with [Babel](https://babeljs.io/) (or similar).
```
eliangcs: Maybe we should rewrite

> but your code has to run on Node `LAMBDA_VERSION`

into something like

> but your code must be runnable on Node `LAMBDA_VERSION`

src/utils/build.js
xavdid: ah, great catch! this isn't honestly that important, I just got annoyed with eslint errors that are supposed to fail showing up and muddying up the test.

README-source.md
```diff
@@ -56,18 +56,14 @@ From a user perspective, both the CLI and the existing web builder platform offe
 
 All Zapier CLI apps are run using Node.js `LAMBDA_VERSION`.
 
-You can develop using any version of Node you'd like, but your code has to run on Node `LAMBDA_VERSION`. You can accomplish this by developing on your preferred version and then transpiling with [Babel](https://babeljs.io/) (or similar).
+You can develop using any version of Node you'd like, but your eventual code must be compatible with `LAMBDA_VERSION`. If you're using features not yet available in `LAMBDA_VERSION`, you can transpile your code to a compatible format with [Babel](https://babeljs.io/) (or similar).
```
eliangcs: 💯

xavdid:
Thanks for the comments! I updated the phrasing.

As for the code location, you're right. The main issue is that `zapier build` doesn't actually run `zapier validate`, it calls the `appWrapper` which validates the schema. So if there's any extra validation (such as this script), it wouldn't happen when the app actually gets built. then we depend on devs running `validate` themselves (which they won't do). The other issue is the paths. We might be able to move that whole thing into a util file and call it in both validate and build? I'll see how hard that is.

PR #238, opened 2018-02-14
be more specific when we cry invalid install

xavdid:
Fixes [PDE-69](https://zapierorg.atlassian.net/browse/PDE-69). the only part of this I wasn't sure on was the purpose of the following [blurb](https://github.com/zapier/zapier-platform-cli/blob/master/src/utils/misc.js#L107-L111) from @bryanhelmig:

```js
// try skipping the CLI itself
const CLIpackageJson = require(path.join(__dirname, '../../package.json'));
if (_.isEqual(packageJson, CLIpackageJson)) {
  return true;
}
```

As far as I can tell, this isn't needed anymore (it's only true if the command is run from within the `platform-cli` directory itself?

Past that, just some nicer messages!

### DIFF
--- a/src/utils/misc.js
+++ b/src/utils/misc.js
@@ -5,6 +5,7 @@ const colors = require('colors/safe');
 const path = require('path');
 const fse = require('fs-extra');
 const os = require('os');
+const semver = require('semver');
 
 const { PLATFORM_PACKAGE } = require('../constants');
 
@@ -89,7 +90,7 @@ const isValidNodeVersion = () => {
 
 const isValidAppInstall = command => {
   if (['help', 'init', 'login', 'apps', 'convert'].includes(command)) {
-    return true;
+    return { valid: true };
   }
 
   let packageJson;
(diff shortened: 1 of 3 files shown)


src/utils/misc.js
```diff
@@ -89,39 +90,46 @@ const isValidNodeVersion = () => {
...
   try {
     packageJson = require(path.join(process.cwd(), 'package.json'));
     const coreVersion = packageJson.dependencies[PLATFORM_PACKAGE];
     // could check for a lot more, but this is probably enough: https://docs.npmjs.com/files/package.json#dependencies
-    if (!coreVersion || coreVersion.includes('^')) {
-      return false;
+    if (!coreVersion) {
+      return {
+        valid: false,
+        reason: `Your app must depend on ${PLATFORM_PACKAGE}`
```
eliangcs: We should provide an actionable info to the users. I'd say something like

> Your app must depend on ${PLATFORM_PACKAGE}. Add ${PLATFORM_PACKAGE} to your package.json as a dependency.

src/utils/misc.js
```diff
@@ -89,39 +90,46 @@ const isValidNodeVersion = () => {
...
+      };
+    } else if (!semver.valid(coreVersion)) {
+      // semver.valid only matches single versions
+      return {
+        valid: false,
+        reason: `Your app must depend on an exact version of ${PLATFORM_PACKAGE}. Found "${coreVersion}" instead`
```
eliangcs: Same here. I'd suggest something like

> Your app must depend on an exact version of ${PLATFORM_PACKAGE}. Found "${coreVersion} in your package.json instead. Try specifying an exact version."

src/utils/misc.js
```diff
@@ -89,39 +90,46 @@ const isValidNodeVersion = () => {
...
   }
 
   try {
-    require(`${process.cwd()}/node_modules/${PLATFORM_PACKAGE}`);
+    require(path.join(process.cwd(), 'node_modules', PLATFORM_PACKAGE));
   } catch (err) {
-    return false;
+    return {
+      valid: false,
+      reason: `Unable to find contents of ${PLATFORM_PACKAGE}, run \`npm install\``
```
eliangcs: I'd suggest something like:

> Looks like you haven't installed ${PLATFORM_PACKAGE}. Try running `npm install`.

eliangcs:
I'm not sure about [that block](https://github.com/zapier/zapier-platform-cli/blob/4cd689b602a499f490f6a7d29b14f5795e8516ed/src/utils/misc.js#L107-L111) of code, either. Looks like we can remove it, @bryanhelmig?

bryanhelmig:
I think it was to prevent either some other install directory or the cli running in its own source directory? If there are no obvious issues present after removing it, 💥 !

PR #240, opened 2018-02-14
zapier convert: Add .environment file and instruction to edit it

eliangcs:
This PR generates a `.environment` file incluing all the auth fields and prints a message instructing users to edit `.environment` when the conversion is done.

### DIFF
--- a/src/commands/convert.js
+++ b/src/commands/convert.js
@@ -33,8 +33,12 @@ const convert = (context, appid, location) => {
   };
 
   return utils.initApp(context, location, createApp).then(() => {
+    context.line();
     context.line(
-      '\nFinished! You might need to `npm install` then try `zapier test`!'
+      `Finished! You may try \`npm install\` and then \`zapier test\` in "${location}" directory.`
+    );
+    context.line(
+      "Also, if your app has authentication, don't forget to edit the environment variables in the .environment file."
     );
   });
 };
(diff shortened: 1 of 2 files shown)


eliangcs:
@xavdid I agree that it's nicer to follow the mainstream convention and use `.env`, but I don't think it deserves a breaking change. We can make it backward compatible by keeping supporting `.environment` (but shows a warning) when `.env` is absent, so we don't have to be forced to include it in our next release.

xavdid:
I'm not quite following, why would we show a warning? If a dev uses `.environment`, they would expect a silent run. 

Anyway I think a major release is a perfect time to change this. We're using something non-standard, so updating the behavior with a better default makes sense. As long as we're clear in the notes, I wouldn't worry about it. If you're calling it without args, your tests will fail. Easy fix though (rename file or pass are). 

Alternatively, we could check for both files? Or overlap for a major and then remove `.environment` on the next?

Either way, i think adhering to industry standards is a worthwhile breaking change

eliangcs:
My intent is not to expand the planned scope of our next release (5.0.0) because I think we already have enough todos in our task queue. So I'm suggesting **NOT** to include the `.environment` -> `.env` change in the next release (5.0.0). Rather, we make it backward compatible by checking both files: 1) attempt to load `.env` first and 2) load `.environment` if `.env` isn't found. Since it's not a breaking change anymore, we can release it in any non-major versions we want, say 5.1.0.

xavdid:
Got it! Sounds good to me.

PR #244, opened 2018-02-20
Version 5

xavdid:
I mixed up how we do changelog a bit (smaller links, a couple of emoji). If you hate it, I can tweak but I thought it might be nice if the notes were a little more human. 

Also considered adding a little blurb saying some highlights, but didn't know if we needed it.


CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
+## 5.0.0
+
+* :exclamation: (improvement, **breaking**) `samples` is now a required field in `BasicOperationSchema` and friends. See [this doc](https://zapier.com/developer/documentation/v2/trigger-sample-results/) for more info ([PR](https://github.com/zapier/zapier-platform-schema/pull/32))
+* :exclamation: (improvement, **breaking**) `order` has been removed from `BasicDisplaySchema` ([PR](https://github.com/zapier/zapier-platform-schema/pull/27))
+
```
eliangcs: [We're using our CHANGELOG.md as an example](https://github.com/zapier/zapier-platform-cli/blob/ac2d65d9ab126e0c710faaf500febff23c665325/src/commands/promote.js#L46) showing users how to write changelog when they do `zapier promote`. Although we aren't parsing the CLI repo's `CHANGELOG.md`, I think it's nice to show users what changelog format we're expecting when they do `zapier promote`. I'm not sure if the blank line here (and the highlight section you considered) would break with our current [changelog parsing logic](https://github.com/zapier/zapier-platform-cli/blob/ac2d65d9ab126e0c710faaf500febff23c665325/src/utils/changelog.js#L6-L40). Could you try that? Maybe adding a test [here](https://github.com/zapier/zapier-platform-cli/blob/ac2d65d9ab126e0c710faaf500febff23c665325/src/tests/utils/changelog.js) is enough.

CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
...
+* :tada: (new) The CLI can now run, test, and build on any version of Node.js! ([PR](https://github.com/zapier/zapier-platform-cli/pull/234))
+* :tada: (new) Read deploy key from env if available ([PR](https://github.com/zapier/zapier-platform-cli/pull/239))
+
+* (improvement) more specific error messages for invalid installs ([PR](https://github.com/zapier/zapier-platform-cli/pull/238))
```
eliangcs: Maybe it's me being picky, but let's be consistent and capitalize the first letter in every sentence, i.e.,

> (improvement) **M**ore specific error messages for invalid installs

.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: node_js
 node_js:
-  - "6.10.2"
+  - "6.10.3"
+  - "7"
+  - "8"
```
eliangcs: If we do this, we have to make sure npm deployment only happen once on a specific Node version. Otherwise, let's not add Node 7 and 8 builds to this PR yet.

xavdid: Great call! This is not how I expected this feature to work. I've updated it per [this](https://docs.travis-ci.com/user/deployment/#Examples-of-Conditional-Deployment) example.

CHANGELOG.md
xavdid: Ah! So I think we shouldn't rely on the exact formatting (number of lines, etc) to generate a changelog. Especially for something as fluid as Markdown, I don't think spacing is a sound assumption to make. Instead, we should develop a simple changelog json schema. We could then parse that json (if it exists) into markdown. 

I think for the purposes of **this release**, we point that suggestion link to a gist containing markdown in the exact spec we want. That way we can do whatever we want for our release notes (since nothing parses those).

For a **future release**, we change this code to parse a schema. We could even have `zapier validate` check it if we wanted to be serious. Offhand, I was thinking something like: 

```
{
  "versions": [
    {
      "version": "1.0.0",
      "notes": [
        "new thing",
        "other new thing"
      ]
    }
  ]
}
```

which would be easy to turn into a standard blurb. We could even get fancy and have `notes` be an array of "update" objects that have type `[bug|improvement|ux]` or something, but that's getting ahead of ourselves.

eliangcs: I'm not sure about writing changelog in JSON, because it was me, I wouldn't write JSON if I can write it in markdown. But it's something we can discuss further. For this release, though, I agree replacing it with a gist link is good enough. 👍

PR #245, opened 2018-02-20
zapier convert: Add track info in converted app's package.json

eliangcs:
Adds to custom fields to the converted app's `package.json`:

* `convertedFromAppID`: the source WB app which was being converted to CLI
* `convertedByCLIVersion`: the CLI version that was used to perform `zapier convert`

The two fields are nested in a field named `zapier`, which allows us to add any other custom fields in the future.

### DIFF
--- a/src/utils/convert.js
+++ b/src/utils/convert.js
@@ -5,6 +5,7 @@ const stripComments = require('strip-comments');
 const { camelCase, snakeCase } = require('./misc');
 const { copyFile, readFile, writeFile, ensureDir } = require('./files');
 const { printStarting, printDone } = require('./display');
+const CLI_VERSION = require('../../package.json').version;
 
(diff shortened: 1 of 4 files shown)


src/utils/convert.js
```diff
@@ -5,6 +5,7 @@ const stripComments = require('strip-comments');
 const { camelCase, snakeCase } = require('./misc');
 const { copyFile, readFile, writeFile, ensureDir } = require('./files');
 const { printStarting, printDone } = require('./display');
+const CLI_VERSION = require('../../package.json').version;
```
xavdid: can import from constants: https://github.com/zapier/zapier-platform-cli/blob/master/src/constants.js#L26

eliangcs:
@xavdid pushed 6d8014e adding a test and importing CLI version from `constants` instead. Also fixed a bug on the way: `zapier convert` crashes when the app description contains a single quote.

PR #254, opened 2018-02-27
Remove 'debug' option from listLog opts

codebycaleb:
Requests to /api/platform/cli/apps/{id}/logs?debug=true result in an error complaining about the 'debug' parameter. This commit will remove any 'debug' opt before sending the request to the logs api.

I'm not a JavaScript expert, so I went to the internet for help regarding cloning objects and removing properties from objects. I hope I did this the way others would (I didn't want to delete the 'debug' opt from the original object; I just wanted to remove it before sending the api request).

### DIFF
--- a/src/utils/api.js
+++ b/src/utils/api.js
@@ -228,9 +228,7 @@ const listInvitees = () => {
 };
 
 const listLogs = opts => {
-  const cleaned = Object.assign({}, opts); // duplicate the original opts
-  delete cleaned.debug; // deletes debug flag if it exists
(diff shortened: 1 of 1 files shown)


codebycaleb:
@xavdid Thanks for that! `lodash` looks awesome – it was even listed in the StackOverflow I originally referenced for removing a key immutably, but I didn't think to check if we used it. I've updated my code accordingly. Thanks for the snippit too!

xavdid:
cool, feel free to merge at your leisure! it'll get deployed when the next version goes out, whenever that is.

PR #260, opened 2018-03-02
[MAJOR] json format only outputs valid json

### LINKED ISSUE #247: Commands that allow a "--format" argument return invalid JSON
The ability to define the display format  of "raw" or "json", for certain commands, has diminished value because there is always a context line at the beginning of what is returned, which makes it technically invalid JSON data. It would be nice to be able to parse and filter this data with tools like [jq](https://stedolan.github.io/jq/) to inform other commands, in for instance a CI pipeline, without having to strip that first line out of the returned string. I propose that this context line only be output when requesting the data as the default "table" format OR to supply another argument to not render the context line.

**Example of problem**

_Command_
```
zapier versions --format=json
```
_Returns_
```
All versions of your app "Example" listed below. <--- This makes the output invalid JSON

[
  {
    "Version": "1.0.41",
    "Platform": "5.0.0",
    "Users": "5",
    "Deployment": "non-production",
    "Deprecation Date": "null",
    "Timestamp": "2018-02-20T14:55:18"
  },
 {
    "Version": "1.0.40",
    "Platform": "5.0.0",
    "Users": "0",
    "Deployment": "non-production",
    "Deprecation Date": "null",
    "Timestamp": "2018-02-20T14:55:18"
  }
]
```

**Solution Example**

_Command_
```
zapier versions --format=json --without-context
```

_Returns_
```
[
  {
    "Version": "1.0.41",
    "Platform": "5.0.0",
    "Users": "5",
    "Deployment": "non-production",
    "Deprecation Date": "null",
    "Timestamp": "2018-02-20T14:55:18"
  },
  {
    "Version": "1.0.40",
    "Platform": "5.0.0",
    "Users": "0",
    "Deployment": "non-production",
    "Deprecation Date": "null",
    "Timestamp": "2018-02-20T14:55:18"
  }
]
```

xavdid:
In order for command output to be piped into other commands (such as `jq`, a json formatter), it can't print extra helpful lines or other, non-json text.

The fix is twofold. The easier side is not printing "these are your apps:" and similar. that's a single check in the `context.line` function.

The more complex half is our custom spinner implementation, which prints out hidden characters when it starts and stops (not to mention, it's very complex), invalidating the output. Moved the spinner to an external lib that's well maintained (`ora`). initial results seem [awesome](https://cdn.zapier.com/storage/photos/e102adcf307a1b02d167601baf1aeda5.png) but there's still a lot of tire-kicking to do there. 

fixes #247 and [PDE-89](https://zapierorg.atlassian.net/browse/PDE-89)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1600,11 +1600,10 @@
       }
     },
     "cli-spinners": {
-      "version": "0.1.2",
+      "version": "1.1.0",
       "resolved":
-        "https://registry.npmjs.org/cli-spinners/-/cli-spinners-0.1.2.tgz",
-      "integrity": "sha1-u3ZNiOGF+54eaiofGXcjGPYF4xw=",
-      "dev": true
+        "https://registry.npmjs.org/cli-spinners/-/cli-spinners-1.1.0.tgz",
+      "integrity": "sha1-8YR7FohE2RemceudFH499JfJDQY="
     },
     "cli-table2": {
       "version": "0.2.0",
@@ -1648,6 +1647,11 @@
       "version": "2.2.0",
       "resolved": "https://registry.npmjs.org/cli-width/-/cli-width-2.2.0.tgz",
       "integrity": "sha1-/xnt6Kml5XkyQUewwR8PvLq+1jk="
+    },
+    "clone": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/clone/-/clone-1.0.3.tgz",
+      "integrity": "sha1-KY1+IjFmD0DAA8LtMUDezz9TCF8="
     },
     "co": {
       "version": "4.6.0",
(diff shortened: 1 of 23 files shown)


xavdid:
We now have a new spinner and most of the commands that have a table can output data as a pure json structure. To test, pipe the output of a command into `jq`: `zapier versions --format=raw | jq`, which should not error. 

Internally, the spinner API is much simpler. All we use now is `startSpinner` and `endSpinner`, which takes a message and a `success` boolean, respectively. The latter also takes a message, but that replaces the started message, which is pretty confusing; i'm thinking of removing it as we don't use it anywhere. We removed a bunch of calls to `clearSpinner()`, since we're no longer worried about things left behind inappropriately, and we don't have to worry about the spinner being there after the cli is exited.

I also removed a param in `callAPI` which we no longer need (because individual commands can choose how to handle errors before bubbling them up to `entry`). 

Notes: 
* all of the commands in [this comment](https://github.com/zapier/zapier-platform-cli/issues/247#issuecomment-369831028) **besides** `describe` and `validate`. They both output multiple tables, so if we want those to be real json, we'll need to change the actual structure of the method. We can do that, but it seems out of scope for what people will probably need this functionality for.
* through some black magic, the spinner running and showing messages doesn't affect the validity of the json output. to see this in action, run `zapier logs --format=raw | jq`, which has a spinner on the screen that succeeds, then outputs valid json anyway. neat!

xavdid:
Sweet! So the downside here is that this is a breaking change, which would require a 6.0 release. I'd love to do that only for CLI and not core/schema, per our discussion.

eliangcs:
Why a breaking change? Is it because we're not printing a context line like "these are your apps:" anymore?

ibolmo:
I'm all in favor of just incrementing major. Our public contract is SEMVER, and it's just easier to adhere to it, then to break it.

PR #269, opened 2018-03-13
zapier convert: Fix output fields for create and search

eliangcs:
#266 added `sample` to trigger/create/search templates and renamed `SAMPLE` to `SAMPLE_FIELDS` for trigger template. But I forgot to rename `SAMPLE` to `SAMPLE_FIELDS` as well for create and search templates. This PR fixes it.

### DIFF
--- a/scaffold/convert/create.template.js
+++ b/scaffold/convert/create.template.js
@@ -356,7 +356,7 @@ module.exports = {
       getInputFields<% } %>
     ],
     outputFields: [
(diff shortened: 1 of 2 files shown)


xavdid:
looks like tests are failing because `SAMPLE_FIELDS` is undefined?

eliangcs:
@xavdid oops, it was a typo. Fixed in 1810f0d.

PR #271, opened 2018-03-14
zapier convert: Call beforeRequest middlewares explicitly for full scripting methods

eliangcs:
In a WB app, we allow users to make HTTP requests with `z.request` themselves in "full" scripting methods like `KEY_poll`, `KEY_write`, and `KEY_search`. However, in a converted CLI app, the `z.request` used in `scripting.js` is not patched to apply `beforeRequest` middlewares. Without the middlewares, authentication won't work.

I spot all the full scripting methods and apply the `beforeRequest` middlewares by calling the newly-added function `runBeforeMiddlewares`.

This PR **MUST** go along with zapier/zapier-platform-legacy-scripting-runner#1.

To test, you can try converting a WB app that has full scripting methods and requires auth. Some example apps:

* 80182 - trigger scripting and session auth
* 82250 - search scripting and OAuth
* 88339 - create scripting and OAuth

You may test like this:

```
zapier convert 80182 trigger-session-auth
cd trigger-session-auth
npm install
npm link zapier-platform-legacy-scripting-runner
echo "ACCESS_TOKEN=a_token" > .environment
echo "SESSION_KEY=secret" >> .environment
zapier test
```

### DIFF
--- a/scaffold/convert/utils.template.js
+++ b/scaffold/convert/utils.template.js
@@ -11,11 +11,6 @@ const replaceVars = (templateString, bundle, result) => {
   return _.template(templateString, options)(values);
 };
 
-// Check if an object is a Promise
-const isPromise = obj => {
-  return typeof obj === 'object' && typeof obj.then === 'function';
-};
-
 // Explicitly run App.beforeRequest middlewares in the app code. Only necessary
 // for WB scripting methods that send HTTP requests themselves, such as
 // KEY_poll, KEY_search, KEY_write, KEY_read_resource, KEY_custom_action_fields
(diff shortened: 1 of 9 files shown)


scaffold/convert/utils.template.js
```diff
@@ -11,6 +11,30 @@ const replaceVars = (templateString, bundle, result) => {
...
+// Explicitly run App.beforeRequest middlewares in the app code. Only necessary
+// for WB scripting methods that send HTTP requests themselves, such as
+// KEY_poll, KEY_search, KEY_write, KEY_read_resource, KEY_custom_action_fields
+// and KEY_custom_search_fields.
+const runBeforeMiddlewares = (request, z, bundle) => {
+  const app = require('./');
+
+  const befores = app.beforeRequest ? app.beforeRequest : [];
+  return befores.reduce((prevResult, before) => {
+    if (!isPromise(prevResult)) {
```
xavdid: Minor, but I think you can skip the check and always wrap it in `Promise.resolve`.

```js
Promise.resolve(Promise.resolve(1)).then(i => console.log(i))
```

PR #273, opened 2018-03-15
zapier convert: Be defensive about request.headers and request.params in middlewares

eliangcs:
There are some occasions where the `request` object, which is passed into a `beforeRequest` middleware, doesn't have `.headers` or `.params` defined, especially when applying middlewares in app code with `runBeforeMiddlewares`. Let's be defensive about it in middlewares by adding `request.headers = request.headers || {}` or `request.params = request.params || {}` everywhere fits.


eliangcs:
@xavdid that's a fair point! I moved the defensive code just before we run the middlewares in 36de625.

PR #274, opened 2018-03-15
zapier convert: Respect auth field keys in test code

eliangcs:
Currently, `zapier convert` always generates the following auth data in test code for API-key-authed WB apps:

```js
const bundle = {
  authData: {
      apiKey: process.env.API_KEY
  },
  // ...
}
```

This PR improves it so that the generated code respect the user's auth field key. So for example, if the auth field key is "my_api_key", it'll generate:

```js
const bundle = {
  authData: {
      my_api_key: process.env.MY_API_KEY
  },
  // ...
}
```

### DIFF
--- a/src/utils/convert.js
+++ b/src/utils/convert.js
@@ -766,6 +766,8 @@ const renderAuthData = definition => {
   const authType = getAuthType(definition);
   let result;
   switch (authType) {
+    case 'api-header': // fall through
+    case 'api-query': // fall through
     case 'basic': {
       let lines = _.map(definition.auth_fields, (field, key) => {
         const upperKey = key.toUpperCase();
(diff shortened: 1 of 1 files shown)


eliangcs:
@xavdid do you mean where the user enters the auth field key? It's in the Web Builder:

screen shot 2018-03-16 at 12 26 01

Let me know if I misunderstood.

xavdid:
In the converter code, where does `my_api_key` get put into the result? That's the piece I'm missing

eliangcs:
@xavdid it's in the generated test code. For example, try running `zapier convert 83342 myapp` and you'll see the `authData` in one of the tests, such as `test/creates/full_create.js`.

xavdid:
Ah, got it. Great!

PR #275, opened 2018-03-20
read from config file first, then write into it

xavdid:
addresses [PDE-105](https://zapierorg.atlassian.net/browse/PDE-105). 

Asking for review from CH, but @BrunoBernardino if you want to start to get spun up on reading code again. 😀


src/utils/api.js
```diff
@@ -137,13 +137,16 @@ const writeLinkedAppConfig = (app, appDir) => {
...
-    prettyJSONstringify({
-      id: app.id,
-      key: app.key
+  // read contents of existing config before writing
+  return readFile(file)
+    .then(configBuff => {
+      return Promise.resolve(JSON.parse(configBuff.toString()));
     })
-  );
+    .catch(() => Promise.resolve({}))
```
eliangcs: Super minor. I'm not familiar with the convention most JS people use, so I may be wrong. Can we compact the code by removing the unnecessary `Promise.resolve()`?

xavdid: Yeah! It's all about control flow. If anything in a promise chain encounters an error, it immediately jumps out and enters the first catch block, right? Typically those are put at the end of the chain, but in our case, we want to catch an error part way and act on it. 

Since the section after this assumes we'll have a JS object, we have to handle errors that happen in the meantime. `readFile` throws an error if the file is gone and `JSON.parse` will fail if it's not json. The catch grabs errors with all of those, ignores the error, and pretends there was blank json in there. This handles first time setup, malformed file, etc.

It's basically the promise equivalent of the following: 

```js
let obj
try {
  const configBuff = await readFile(file)
  obj = JSON.parse(configBuff.toString())
} catch {
  obj = {} // Promise.resolve({})
}

Object.assign({}, obj, { id: app.id, key: app.key })
```

package.json
```diff
@@ -72,6 +72,7 @@
     "litdoc": "1.5.6",
     "markdown-toc": "1.1.0",
     "mocha": "3.4.2",
+    "mock-fs": "^4.4.2",
```
BrunoBernardino: Can we specify a specific version instead of `^`? I know with `package-lock.json` this is less of an issue, but just to keep it consistent with how we set the other dependencies.

src/utils/api.js
BrunoBernardino: This makes sense and I think we can let it be. I'd just add a comment saying something like `// We don't want this to error out. It's possible the file doesn't exist.`

The only main thing that could be an issue would be bad permissions, which would be ignored here, but caught in `writeFile` after.

xavdid:
Cool, fixed in bb63eec. Thanks y'all!

PR #280, opened 2018-03-23
Add rawError to callAPI

xavdid:
addresses [PDE-138](https://zapierorg.atlassian.net/browse/PDE-138)

Previously, when the server returned a 400+ error code, we would package the error message up and throw it. This is normally useful, but when we get back a json with errors and want to process them, we were (previously) unable to. It also more correctly identifies when then the issue is an activation one.

This adds a param to `callAPI` that returns the response object instead of directly throwing an error. This is most relevant in `promote`, but it will potentially be useful in other places as well. While I was there, I make some of the language more clear too.

Before: 

After:

**Merging Notes**
* This depends on https://github.com/zapier/zapier/pull/16441, so it may not get merged right away.
* this is going to cause a messy merge with #247, I'll handle that when we get to it.

### DIFF
--- a/src/commands/promote.js
+++ b/src/commands/promote.js
@@ -1,5 +1,6 @@
 const colors = require('colors/safe');
 const utils = require('../utils');
+const _ = require('lodash');
 
 const hasCancelled = answer =>
   answer.toLowerCase() === 'n' ||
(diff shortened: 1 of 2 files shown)


xavdid:
I thought about it, but since that's clearly an example it sort of makes sense to be there. Or if you're more concerned with them matching, i'm happy to. 

It would be cool to get some system for keeping the help text and command output in sync. Since the latter happens at runtime it's probably not possible, but it would be useful.

eliangcs:
@xavdid since the help text is an example output of the command, I thought it'd better if we can keep them in sync. Just #fyi, not a blocker.

PR #281, opened 2018-03-23
zapier convert: Fix session auth environment variable name

eliangcs:
Addresses [PDE-71](https://zapierorg.atlassian.net/browse/PDE-71).

Fixes two issues:

* When converting session auth, `.env` had user-facing auth fields like `USERNAME` and `PASSWORD` while it should have `SESSION_KEY`. (Because `zapier test` doesn't actually test session auth refreshing.)
* Use `_.snakeCase(key).toUpperCase()` instead of `key.toUpperCase()` to turn an auth field key to the environment variable name, so before the generated test code has this:

```js
authData: {
  sessionKey: process.env.SESSIONKEY
}
```

Now it becomes:

```js
authData: {
  sessionKey: process.env.SESSION_KEY
}
```

### DIFF
--- a/src/utils/convert.js
+++ b/src/utils/convert.js
@@ -763,46 +763,46 @@ const writeStep = (type, definition, key, legacyApp, newAppDir) => {
   );
 };
 
-// render the authData used in the trigger/search/create test code
-const renderAuthData = definition => {
+// Get auth field keys that will be put into test code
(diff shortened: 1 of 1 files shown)


eliangcs:
@BrunoBernardino will make sure to do that next time. Thanks for the review and suggestion!

PR #282, opened 2018-03-24
Less Nagging

xavdid:
Per our conversation, pull out some of the reminding about updating version. 

I pulled the custom update method because the user needs to update both CLI and the core dep in each of their apps, which gets a little wordy for an update message. Maybe we should add a section to the doc about what you need to update and when?

### DIFF
--- a/README-source.md
+++ b/README-source.md
@@ -1366,39 +1366,34 @@ source ~/.bash_completion.d/_zapier
 
 Finally, restart your shell and start hitting TAB with the `zapier` command!
 
-## Upgrading Zapier Platform CLI or Zapier Platform Core
+## The Zapier Platform Packages
 
-The CLI version of the Platform has 3 public packages:
+The Zapier Platform consists of 3 npm packages that are released simultaneously as a trio.
 
-- [`zapier-platform-cli`](https://github.com/zapier/zapier-platform-cli): This is the `zapier` command. Should be installed with `npm install -g zapier-platform-cli` and is used to interact with your app and Zapier, but doesn't run in nor is included by your app.
+- [`zapier-platform-cli`](https://github.com/zapier/zapier-platform-cli) is the code that powers the `zapier` command. You use it most commaonly with the `test`, `scaffold`, and `push` commands. It's installed with `npm install -g zapier-platform-cli` and does not correspond to a particular app.
 
-- [`zapier-platform-core`](https://github.com/zapier/zapier-platform-core): This is what allows your app to interact with Zapier. Apps require this package in their `package.json` file, and a specific version by design, so you can keep using an older "core" version without having to upgrade your code, but still being able to update your app. Should be installed with `npm install` in your app. You should manually keep this version number in sync with your `zapier-platform-cli` global one, but it's not required for all upgrades.
+- [`zapier-platform-core`](https://github.com/zapier/zapier-platform-core) is what allows your app to interact with Zapier. It holds the `z` object and app tester code. Your app depends on a specific version of `zapier-platform-core` in the `package.json` file. It's installed via `npm install` along with the rest of your app's dependencies.
 
-- [`zapier-platform-schema`](https://github.com/zapier/zapier-platform-schema): This is a separate package for maintainability purposes. It's a dependency installed automatically by `zapier-platform-core`. It's basically the package that enforces app schemas to match what Zapier has in the backend.
+- [`zapier-platform-schema`](https://github.com/zapier/zapier-platform-schema) enforces app structure behind the scenes. It's a dependency of `core`, so it will be installed automatically.
 
-### Upgrading Zapier Platform CLI
+### Updating
 
-Check your version with `$ zapier --version` and update with `$ npm -g update zapier-platform-cli`.
+The Zapier platform and its tools are under active development While you don't need to install every release, we release new versions because they are better than the last. We do our best to adhere to [Semantic Versioning](https://semver.org/) wherein we won't break your code unless there's a `major` release. Otherwise, we're just fixing bugs (`patch`) and adding features (`minor`).
 
-You can also [look at the Changelog](https://github.com/zapier/zapier-platform-cli/blob/master/CHANGELOG.md) to understand what changed between versions.
+Barring unforseen circumstances, all released platform versions will continue to work for the forseeable future. While you never *have* to upgrade your app's platform version, we recommend keeping an eye on the [changelog](https://github.com/zapier/zapier-platform-cli/blob/master/CHANGELOG.md) to see what new features and bux fixes are available.
 
-#### When to Update CLI
+<!-- TODO: if we decouple releases, change this -->
+The most recently released version of `cli` and `core` is `PACKAGE_VERSION`. You can see the versions you're working with by running `zapier -v`.
 
-You should keep your `zapier-platform-cli` global package up-to-date to get the latest features and bug fixes for interacting with your app, but it's not mandatory. You should be able to use any publicly available `zapier-platform-cli` version.
+To update `cli`, run `npm install -g zapier-platform-cli`.
 
-### Upgrading Zapier Platform Core
+To update the version of `core` your app depends on, set the `zapier-platform-core` dependency in your `package.json` to a version listed [here](https://github.com/zapier/zapier-platform-core/releases) and run `npm install`.
 
-Open your app's `package.json` file and update the number for the `zapier-platform-core` entry to what the latest public version is. You can see it on the right side of the [NPM registry page](https://www.npmjs.com/package/zapier-platform-core). After that you should be able to run `npm update` and get the latest goodies!
-
-#### When To Update Core
-
-You should keep your `zapier-platform-core` package up-to-date to get the latest features and bug fixes for your app when interacting with Zapier, but it's not mandatory. You should be able to use any publicly available `zapier-platform-core` version.
-
-While not always required, it's also recommended you use the same `zapier-platform-cli` global package as the app's `zapier-platform-core` one, to ensure maximum compatibility.
+For maximum compatibility, keep the versions of `cli` and `core` in sync.
 
 ## Development of the CLI
 
-This section is only relevant if you're editing the `zapier-platform-cli` package
+This section is only relevant if you're editing the `zapier-platform-cli` package itself.
 
 ### Commands
 
--- a/README-source.md
+++ b/README-source.md
@@ -471,7 +471,7 @@ In some cases, it might be necessary to provide fields that are dynamically gene
 [insert-file:./snippets/custom-fields.js]
 ```
 
-Additionally, if there is a field that affects the generation of dynamic fields, you can set the `altersDynamicFields: true` property. This informs the Zapier UI that whenver the value of that field changes, fields need to be recomputed. An example could be a static dropdown of "dessert type" that will change whether the function that generates dynamic fields includes a field "with sprinkles."
+Additionally, if there is a field that affects the generation of dynamic fields, you can set the `altersDynamicFields: true` property. This informs the Zapier UI that whenever the value of that field changes, fields need to be recomputed. An example could be a static dropdown of "dessert type" that will change whether the function that generates dynamic fields includes a field "with sprinkles."
 
 ```javascript
 [insert-file:./snippets/alters-dynamic-fields.js]
@@ -982,7 +982,7 @@ various kinds of errors occur.
 
 Errors due to a misconfiguration in a user's Zap should be handled in your app
 by throwing a standard JavaScript `Error` with a user-friendly message.
-Typically, this will be prettifying 4xx responses or API's that return errors as
+Typically, this will be prettifying 4xx responses or APIs that return errors as
 200s with a payload that describes the error.
 
 Example: `throw new Error('Your error message.');`
@@ -1370,7 +1370,7 @@ Finally, restart your shell and start hitting TAB with the `zapier` command!
 
 The Zapier Platform consists of 3 npm packages that are released simultaneously as a trio.
 
-- [`zapier-platform-cli`](https://github.com/zapier/zapier-platform-cli) is the code that powers the `zapier` command. You use it most commaonly with the `test`, `scaffold`, and `push` commands. It's installed with `npm install -g zapier-platform-cli` and does not correspond to a particular app.
+- [`zapier-platform-cli`](https://github.com/zapier/zapier-platform-cli) is the code that powers the `zapier` command. You use it most commonly with the `test`, `scaffold`, and `push` commands. It's installed with `npm install -g zapier-platform-cli` and does not correspond to a particular app.
 
 - [`zapier-platform-core`](https://github.com/zapier/zapier-platform-core) is what allows your app to interact with Zapier. It holds the `z` object and app tester code. Your app depends on a specific version of `zapier-platform-core` in the `package.json` file. It's installed via `npm install` along with the rest of your app's dependencies.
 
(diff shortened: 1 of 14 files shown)


xavdid:
Before it happened in upload. That seemed like a weird time to check (maybe because we got all that info back from the server?) so `test` made more sense to me. Could also be `validate`, but then you run into the snafu where the extra code in `zapier validate` doesn't run during build.

eliangcs:
I don't have strong feelings, but there may be some devs not running `zapier test` at all. Maybe we can put it in `zapier build`? @BrunoBernardino what's your opinion?

xavdid:
Totally. The only other piece to mention is that `updateNotifier` only notifies on the second run. That is, the first time it's run, it checks to see if it's out of date and stores the fact that it is. The second time it's run, it notes that fact that it has an update stored and announces it. Here's the stored json:

```
{
	"optOut": false,
	"lastUpdateCheck": 15264 // shortened to manually trigger a check
}
```

```json
{
	"optOut": false,
	"lastUpdateCheck": 1522132259312,
	"update": {
		"latest": "5.0.0",
		"current": "4.3.2",
		"type": "major",
		"name": "zapier-platform-core"
	}
}
```

It's probably fine, but I would be more on the dev running `test` twice than `build`. Probably fine either way though!

src/entry.js
```diff
@@ -46,17 +50,10 @@ module.exports = argv => {
...
-      message: `Update available ${colors.grey(
-        notifier.update.current
-      )} → ${colors.green(notifier.update.latest)}\nRun ${colors.cyan(
-        'npm i -g zapier-platform-cli'
-      )} to update, and then ${colors.blue('re-test your integration')}.`
-    });
-  }
+  updateNotifier({
+    pkg: require('../package.json'),
+    updateCheckInterval: UPDATE_NOTIFICATION_INTERVAL
```
BrunoBernardino: The reason why we added this specific message was to make it clearer what needed to be done by the developer (highlighting the problem, command, and remind to re-test).

It was part of a UX research/review, so I'm not so sure we'd want to remove it.

If we found out something more about this in the meanwhile, then it's perfectly fine to change! #recommendation

src/utils/misc.js
```diff
@@ -151,13 +153,73 @@ const entryPoint = dir => {
   return fse.realpathSync(path.resolve(dir, packageJson.main));
 };
 
+const printVersionInfo = context => {
```
BrunoBernardino: I like this helper, nice job! I think it's acceptable to print the versions while doing a `push`/`upload`.

src/utils/api.js
```diff
@@ -274,18 +273,8 @@ const upload = (zipPath, appDir) => {
         }
       });
     })
-    .then(appVersion => {
+    .then(() => {
       printDone();
```
BrunoBernardino: If I'm reading this correctly, we're no longer running this version check on the `push`/`upload`, only on `version`, and we also no longer provide the URL to help update the dependencies, or understand what they are and why they're different.

Adding the check and URL on these commands was made via UX research/review so unless we found out something more, I don't feel very comfortable removing that as it might re-create confusion. #recommendation

BrunoBernardino:
I reviewed without noticing the comments. `zapier build` they're required to run, `zapier test` they're only encouraged, so we should keep that important check in "mandatory" commands. #recommendation

docs/cli.html
```diff
@@ -948,7 +948,7 @@ <h2 id="logs">logs</h2>
 <li><code>--status={any,success,error}</code> -- <em>optional</em>, display only success logs (status code &lt; 400 / info) or error (status code &gt; 400 / tracebacks). Default is <code>any</code></li>
 <li><code>--type={console,http}</code> -- <em>optional</em>, display only console or http logs. Default is <code>console</code></li>
 <li><code>--detailed</code> -- <em>optional</em>, show detailed logs (like request/response body and headers)</li>
-<li><code>--user=user@example.com</code> -- <em>optional</em>, display only this user&apos;s logs. Default is <code>me</code></li>
+<li><a href="mailto:`--user=user@example.com">`--user=user@example.com</a><code>-- _optional_, display only this user&apos;s logs. Default is</code>me`</li>
```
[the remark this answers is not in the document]
xavdid: not sure! I didn't change this, but it popped out in the docs. I just did a `rm -rf node_modules && npm i && npm run docs` and had no local changes on this branch. the `package-lock` should guarantee that others have the same result.

src/entry.js
xavdid: I like the clear instructions element of it, but  I felt compelled to change it because the message was misleading. It said to "run `npm i -g zapier-platform-cli` and then re-test your integration". Since the test code will break when `core` is updated (not cli), re-running `test` there doesn't reflect the new version. A better message might describe updating the dep in `package.json` and then re-running npm install, but that on top of wanting them to update CLI seemed to get really wordy for a quick "pop up". Also, assuming that they're not updating across a major version, there's really no need to re-test. The semver package allows us to get major|minor|patch from a pair of versions (`.difference`) so we could selectively include the re-test message if it's important. 

Was the concern on the original UX finding that the dev wouldn't know how to update the dependency or wasn't sure what effect doing so would have? I feel like I'm missing some context there. Do you know offhand where the result of those test(s) were?

eliangcs:
Bruno gave a really thorough review and I totally agree with his comments. In summary, here are what I have in mind:

* It's 💯 to check update for `cli` in `entry.js` 
* Maybe remove the check for `cli` from `zapier version` (i.e. `printVersionInfo`)?
* I suggest we check update for `core` only in `zapier build` and not elsewhere

src/utils/api.js
xavdid: Sure, so there's two different (similar) pieces of code being run. in `zapier -v`, it previously spit out the node version and CLI version. I expanded that to include the depended-on core version (convenient) and figured I'd expand it to do a check for whether the installed version matches the required version. This is simply informational and only run on demand.

Separately, `test` (but probably soon `build` per discussion here) checks to see if the required version is behind the most recently published version and once a week (when `build` is run) it'll notify them that an update is available. The main goal of this PR was to reduce the amount of times we told them they were out of date (per feedback from [this typeform](https://admin.typeform.com/form/yPC7uY/results#responses)). Especially when the code is running fine, there's no particular need to upgrade and us pushing that aggressively devs (`core` on every upload, daily for `CLI`) was causing friction.

README.md
```diff
@@ -2375,39 +2372,34 @@ source ~/.bash_completion.d/_zapier
 
 Finally, restart your shell and start hitting TAB with the `zapier` command!
 
-## Upgrading Zapier Platform CLI or Zapier Platform Core
+## The Zapier Platform Packages
```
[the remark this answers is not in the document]
xavdid: I'm having a little trouble following what you're recommending here. Is the problem with the highlighted line or the content of the section in general? 

I think part of ~our~ my (don't want to speak for everyone) problem here is that we've worked extensively with these packages (and node in general) so a lot of the necessary actions here seem very obvious. Since CLI and core are updated separately (directly from npm vs editing a file and re-installing) it made more sense to me to thematically break them out a bit. Also thinking about the user with multiple apps, CLI will be updated once but core will be updated repeatedly, so being able to access those separately in the docs made sense.

xavdid:
@eliangcs Sweet. Tomorrow I'll move the `updateNotifier` code from `test` into `build`. Per your other comments, I there might be some confusion between what happens in `printVersionInfo` and the other places. Per [this comment](https://github.com/zapier/zapier-platform-cli/pull/282#discussion_r177354889), the code that has `updateNotifier` is what actually prints the prompt to upgrade to a newer available version (via information gleaned from npm). That (only) happens in `entry` for `cli` and `test` (soon `build`) for `core`. The code in `printVersionInfo` just prints info about what's currently installed/required and doesn't do a check against the server. It's a small expansion on previous `-v` behavior. 

Given that, besides moving code and agreeing on doc content, are there any other action items?

README-source.md
```diff
@@ -1366,39 +1366,34 @@ source ~/.bash_completion.d/_zapier
...
 
-### Upgrading Zapier Platform CLI
+### Updating
 
-Check your version with `$ zapier --version` and update with `$ npm -g update zapier-platform-cli`.
+The Zapier platform and its tools are under active development While you don't need to install every release, we release new versions because they are better than the last. We do our best to adhere to [Semantic Versioning](https://semver.org/) wherein we won't break your code unless there's a `major` release. Otherwise, we're just fixing bugs (`patch`) and adding features (`minor`).
 
-You can also [look at the Changelog](https://github.com/zapier/zapier-platform-cli/blob/master/CHANGELOG.md) to understand what changed between versions.
+Barring unforseen circumstances, all released platform versions will continue to work for the forseeable future. While you never *have* to upgrade your app's platform version, we recommend keeping an eye on the [changelog](https://github.com/zapier/zapier-platform-cli/blob/master/CHANGELOG.md) to see what new features and bux fixes are available.
```
eliangcs: I like how you edited this paragraph. It's much clearer now. But the term "platform version" doesn't seem super clear to me. And this is the first time we see this term in the doc. Maybe we should change it to something like "package version"? So the whole paragraph becomes:

> Barring unforseen circumstances, all released package versions will continue to work for the forseeable future. While you never *have* to upgrade your app's `core` package version, we recommend keeping an eye on the changelog ...

src/entry.js
eliangcs: @xavdid has a point. Maybe we should move the message ("re-test your integration") to where we notify upgrading `core`?

README.md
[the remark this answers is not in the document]
BrunoBernardino: Sorry that wasn't clear! My comment was related to the content of the section in general.

I agree with you on the experience with node, but we've seen developers get confused about `core` vs `cli`, why each existed, and how/where to update each.

If we've found out this isn't the case anymore, or that this new section caused more confusion, I'm OK with it being changed/removed, otherwise I'd prefer to keep it.

docs/cli.html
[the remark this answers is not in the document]
BrunoBernardino: No worries, was just wondering.

src/entry.js
BrunoBernardino: That was a request by @benzapier and @ZaneLyon IIRC. I didn't check where the data for that request came from, but seemed anecdotal.

I agree the message would be quite large for the whole correct instruction, so maybe @eliangcs 's suggestion makes more sense.

src/utils/api.js
BrunoBernardino: Ok, that makes sense. Thank you for the explanation. 🙇 

I think we'd be more inline with those objectives if we are more explicit on showing them how to update what (like a link to a doc, similar to what we had) on `build`.

I'm not entirely sold on the benefits of splitting the `core` vs `cli` check, but we don't really have a lot of data on it, so I'm fine rolling with it, as you've been more involved in the recent research than I have.

README.md
```diff
@@ -2375,39 +2372,34 @@ source ~/.bash_completion.d/_zapier
...
+### Updating
 
-Check your version with `$ zapier --version` and update with `$ npm -g update zapier-platform-cli`.
+The Zapier platform and its tools are under active development While you don't need to install every release, we release new versions because they are better than the last. We do our best to adhere to [Semantic Versioning](https://semver.org/) wherein we won't break your code unless there's a `major` release. Otherwise, we're just fixing bugs (`patch`) and adding features (`minor`).
 
-You can also [look at the Changelog](https://github.com/zapier/zapier-platform-cli/blob/master/CHANGELOG.md) to understand what changed between versions.
+Barring unforseen circumstances, all released platform versions will continue to work for the forseeable future. While you never *have* to upgrade your app's platform version, we recommend keeping an eye on the [changelog](https://github.com/zapier/zapier-platform-cli/blob/master/CHANGELOG.md) to see what new features and bux fixes are available.
 
-#### When to Update CLI
+<!-- TODO: if we decouple releases, change this -->
```
BrunoBernardino: Is this version manually written here? We should have it as a var of sorts, or link to the releases tab, for example, so we don't have to keep updating this manually and potentially forget about it.

README.md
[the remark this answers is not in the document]
xavdid: Nothing to apologize for. 😄The goal here was to clean up the language, the content itself was good.

src/utils/api.js
xavdid: Ah, the reason they need to be in separate spots is because CLI can be run independently of an app. the CLI check can conveniently be put in `entry` which is run every time the CLI is invoked. We have to attach the check for `core` somewhere where we know we're working with an app. Your suggestion of `build` is a good one, since that's probably the most commonly run command.

README-source.md
```diff
@@ -1366,39 +1366,34 @@ source ~/.bash_completion.d/_zapier
...
+- [`zapier-platform-core`](https://github.com/zapier/zapier-platform-core) is what allows your app to interact with Zapier. It holds the `z` object and app tester code. Your app depends on a specific version of `zapier-platform-core` in the `package.json` file. It's installed via `npm install` along with the rest of your app's dependencies.
 
-- [`zapier-platform-schema`](https://github.com/zapier/zapier-platform-schema): This is a separate package for maintainability purposes. It's a dependency installed automatically by `zapier-platform-core`. It's basically the package that enforces app schemas to match what Zapier has in the backend.
+- [`zapier-platform-schema`](https://github.com/zapier/zapier-platform-schema) enforces app structure behind the scenes. It's a dependency of `core`, so it will be installed automatically.
 
-### Upgrading Zapier Platform CLI
+### Updating
```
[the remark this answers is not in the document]
xavdid: Honestly I forget that you're not, your English is great. Update and upgrade are pretty interchangeable. I changed it because npm [uses it](https://docs.npmjs.com/cli/update), but ironically it's aliased as `upgrade`. [This answer](https://english.stackexchange.com/questions/13353/are-there-any-differences-between-update-and-upgrade) also says they're pretty much the same. 

To me, update feels like it's more for it's replacing an older version, upgrade feels like replacing it with a different, better module (which is an upgrade over the last). 

But really, either way.

README-source.md
xavdid: Nice! Changed to 

> never *have* to upgrade your app's `platform-core` dependency

docs/cli.html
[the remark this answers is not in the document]
xavdid: so I dug in here. seems like it's because litdoc (which generates docs) installs a range and last time the package-lock got cleared, we slid up a version that changed the way they parsed emails ([this](https://github.com/markedjs/marked/pull/1034) PR, probably went out in `0.3.14`). I agree it looks a little weird, but probably not worth fixing at this point

src/entry.js
xavdid: Sweet, grabbed the best of both worlds.

xavdid: er, might swap the version numbers to after the filename

xavdid: This is the version i'm pushing:

xavdid:
Ok! Pushed a new version @BrunoBernardino. 

Things to test: 

* run `zapier -v` in root and in an app directory
* edit the file at `~/.config/configstore/update-notifier-zapier-platform-core.json` to have an old timestamp and then run `zapier build` twice in an app directory

src/entry.js
BrunoBernardino: Thanks, that looks awesome!

src/utils/api.js
BrunoBernardino: I see, that sounds like a great plan, then!

BrunoBernardino:
I've checked the changes and they look great! I've got other PT3 duties that would halt me pulling and testing this PR, so @eliangcs will take care of that instead, to ship this sooner.

eliangcs:
@xavdid I pushed some commits to address two bugs I found. You can find the bug description in commit 9e45d10 and f0ce5f0. And 6b0b302 is just a follow-up to f0ce5f0.

Now all looks good! Merging.

PR #283, opened 2018-03-27
Fix issue where calls to utils.printData hangs

### LINKED ISSUE #279: Calls to utils.printData hangs for 'row' format with small displays
When running a command with an optional `format` flag set to `row` or `table` (e.g. `zapier apps --format=row`), if the terminal window is "skinny" (i.e. less than 15px), the command will hang indefinitely.

I did some debugging (via chrome's node debugger) and found the offending area I think:

We get stuck in an infinite loop slicing `rest` when `widthForValue` is non-positive. I'm not entirely familiar with what's going on here but I'm happy to try my hand at fixing this issue – I just need to get a little more familiar with what we're doing here.

## Your Environment

* CLI Version (`zapier -v`): ```zapier-platform-cli/5.0.0
node/v9.5.0```
* Operating System and version: `MacOS High Sierra 10.13.3`

codebycaleb:
Resolves #279 

When calls to `utils.printData` were made on a headless machine (such as a CI) or a "skinny" terminal (with a column width less than a certain amount of columns - 15 would always do it), the call would get stuck in a while loop. To make a long story short, this loop was lopping off text based on how many columns we had calculated were available for use. But when the calculated width was non-positive, the text-lopping never ended because we were saying ~'drop the first -6 characters of "example"'.

Now, when the calculated width is non-positive, the entire while-loop is avoided by directly returning the `makePlain` call.

As a bonus, this should fix an old test that was marked skipped. :)

### DIFF
--- a/src/tests/utils.js
+++ b/src/tests/utils.js
@@ -9,18 +9,62 @@ describe('utils', () => {
     global._babelPolyfill.should.eql(true);
   });
 
-  // For some reason, Travis dies on this one
-  it.skip('should print a nice little table', () => {
-    const table = utils.makeTable(
-      [{ id: 123, title: 'hello' }, { id: 456, title: 'world' }],
-      [['ID', 'id'], ['Title', 'title'], ['Missing', 'missing']]
-    );
-    table.indexOf('ID').should.be.above(0);
-    table.indexOf('Title').should.be.above(0);
-    table.indexOf('Missing').should.be.above(0);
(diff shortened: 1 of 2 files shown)


codebycaleb:
D'oh! I saw the comment but I didn't realize it meant Jira. Got it for the future. Thanks!

xavdid:
No problem at all! Anyway, looks great. go ahead and merge it when you're ready.

PR #288, opened 2018-03-30
Print npm stderr if `npm install` fails

### LINKED ISSUE #82: Windows: zapier push fails because build's node_modules folder is empty
```
C:\Program Files\nodejs\node_modules\npm\zapier-example>zapier push --debug
Preparing to build and upload your app.

>> GET https://zapier.com/api/platform/cli/check
<< 200
<< {"success": true}

  Copying project to temp directory     done!
  Installing project dependencies .

Running npm.cmd install --production command in C:\Users\allen\AppData\Local\Te
p\zapier-536ab35f:

..


    done!
  Applying entry point file ...
Error: : File C:\Users\allen\AppData\Local\Temp\zapier-536ab35f/node_modules/za
ier-platform-core/include/zapierwrapper.js not found.
    at C:\Users\allen\AppData\Roaming\npm\node_modules\zapier-platform-cli\lib\
tils\files.js:31:11

Error!

```
```
C:\Program Files\nodejs\node_modules\npm\zapier-example>npm version
{ 'zapier-platform-example-app-trigger': '1.0.0',
  npm: '4.6.1',
  ares: '1.10.1-DEV',
  http_parser: '2.7.0',
  icu: '57.1',
  modules: '48',
  node: '6.9.1',
  openssl: '1.0.2j',
  uv: '1.9.1',
  v8: '5.1.281.84',
  zlib: '1.2.8' }

```

eliangcs:
Fixes #82 and zapier/zapier#16579.

Our current code in `zapier build ` assumes if `npm install` returns 0, everything is good. But `npm install` may fail without returning a non-zero exit code. For example, try `npm install` when the `name` in `package.json` has a space character. In this case, `npm` prints the error to stderr, but exits with 0 as if it was successful.

So what we should do instead, is to check further if `APP_DIR/node_modules/zapier-platform-core` exists to make sure `npm install` is really successful before moving on to the next step.

With this PR, now if you `zapier build` an app with a space in thier `name` in `package.json`, we'll show the stderr from `npm`:

```
$ zapier build
Building project.

  Copying project to temp directory ⠃ done!
  Installing project dependencies ⠆

Error! Could not install dependencies properly. Error log:
npm WARN Invalid name: "my app"
npm WARN zapier-d1da7977 No description
npm WARN zapier-d1da7977 No repository field.
npm WARN zapier-d1da7977 No README data
npm WARN zapier-d1da7977 No license field.


(Use --debug flag and run this command again to get more details.)
```

### DIFF
--- a/src/utils/build.js
+++ b/src/utils/build.js
@@ -309,6 +309,20 @@ const build = (zipPath, wdir) => {
       printStarting('Installing project dependencies');
       return runCommand('npm', ['install', '--production'], { cwd: tmpDir });
     })
+    .then(output => {
+      // `npm install` may fail silently without returning a non-zero exit code, need to check further here
+      const corePath = path.join(
+        tmpDir,
+        'node_modules',
+        constants.PLATFORM_PACKAGE
+      );
+      if (!fs.exists(corePath)) {
+        throw new Error(
+          'Could not install dependencies properly. Error log:\n' +
+            output.stderr
+        );
+      }
+    })
     .then(() => {
       printDone();
       printStarting('Applying entry point file');
--- a/src/utils/build.js
+++ b/src/utils/build.js
@@ -316,7 +316,7 @@ const build = (zipPath, wdir) => {
         'node_modules',
         constants.PLATFORM_PACKAGE
       );
-      if (!fs.exists(corePath)) {
+      if (!fs.existsSync(corePath)) {
         throw new Error(
           'Could not install dependencies properly. Error log:\n' +
             output.stderr
--- a/src/utils/misc.js
+++ b/src/utils/misc.js
@@ -43,9 +43,10 @@ const runCommand = (command, args, options) => {
     let stdout = '';
     if (result.stdout) {
       result.stdout.on('data', data => {
-        stdout += data.toString();
+        const str = data.toString();
+        stdout += str;
         if (global.argOpts.debug) {
-          console.log(colors.green(stdout));
+          process.stdout.write(colors.green(str));
         }
       });
     }
@@ -53,9 +54,10 @@ const runCommand = (command, args, options) => {
     let stderr = '';
     if (result.stderr) {
       result.stderr.on('data', data => {
-        stderr += data.toString();
+        const str = data.toString();
+        stderr += str;
         if (global.argOpts.debug) {
-          console.log(colors.red(stdout));
+          process.stderr.write(colors.red(str));
         }
       });
     }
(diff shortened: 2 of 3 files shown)


src/utils/misc.js
```diff
@@ -66,7 +68,7 @@ const runCommand = (command, args, options) => {
       if (code !== 0) {
         reject(new Error(stderr));
       }
-      resolve(stdout);
+      resolve({ stdout, stderr });
```
FokkeZB: I assume you've checked we're not expecting this to resolve with only `stdout` anywhere else?

src/utils/misc.js
```diff
@@ -43,19 +43,21 @@ const runCommand = (command, args, options) => {
     let stdout = '';
     if (result.stdout) {
       result.stdout.on('data', data => {
-        stdout += data.toString();
+        const str = data.toString();
+        stdout += str;
         if (global.argOpts.debug) {
-          console.log(colors.green(stdout));
+          process.stdout.write(colors.green(str));
```
FokkeZB: Why the change from `console.log` to `process.stdout`?

src/utils/misc.js
eliangcs: I was supposed to. Thanks for the reminder! Fixed in a218ba4.

src/utils/misc.js
eliangcs: `console.log` appends a newline every call, but the stdout from the external command already has the newlines.

eliangcs:
@FokkeZB thanks for reviewing. I wouldn't worry about adding a test for now though. To test it, we need to test the `build()` function in the `utils/build` module, which requires quite an amount of effort as there's no test like that at the moment. Doesn't seem worthwhile for a UX/DX enhancement.

src/utils/build.js
```diff
@@ -309,6 +309,20 @@ const build = (zipPath, wdir) => {
...
       return runCommand('npm', ['install', '--production'], { cwd: tmpDir });
     })
+    .then(output => {
+      // `npm install` may fail silently without returning a non-zero exit code, need to check further here
+      const corePath = path.join(
+        tmpDir,
+        'node_modules',
+        constants.PLATFORM_PACKAGE
+      );
+      if (!fs.exists(corePath)) {
```
xavdid: `fs.exists` is both asynchronous and deprecated ([docs](https://nodejs.org/docs/latest-v6.x/api/fs.html#fs_fs_exists_path_callback)). I believe this needs to be `fs.existsSync` or one of the other methods mentioned in that doc link

eliangcs: Nice catch! Fixed in 5fb57c1.

PR #301, opened 2018-04-17
Ignoring missing package for browserify bundling.

### LINKED ISSUE #189: zapier push fails with "Cannot find module 'supports-color'"
When running `zapier push`, I get the following error : 
```
Preparing to build and upload your app.

  Copying project to temp directory ⠃ done!
  Installing project dependencies ⠃ done!
  Applying entry point file ⠃ done!
  Building app definition.json ⠃ done!
  Validating project ⠃ done!
  Zipping project and dependencies ⠰events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: Cannot find module 'supports-color' from '/private/var/folders/dw/d_f_yf1x5zz2fm1rtkpxzmw00000gn/T/zapier-19a7602e/node_modules/debug/src'
    at /Users/aymericbouzy/.nvm/versions/node/v6.10.2/lib/node_modules/zapier-platform-cli/node_modules/resolve/lib/async.js:48:31
    at processDirs (/Users/aymericbouzy/.nvm/versions/node/v6.10.2/lib/node_modules/zapier-platform-cli/node_modules/resolve/lib/async.js:182:39)
    at ondir (/Users/aymericbouzy/.nvm/versions/node/v6.10.2/lib/node_modules/zapier-platform-cli/node_modules/resolve/lib/async.js:197:13)
    at load (/Users/aymericbouzy/.nvm/versions/node/v6.10.2/lib/node_modules/zapier-platform-cli/node_modules/resolve/lib/async.js:80:43)
    at onex (/Users/aymericbouzy/.nvm/versions/node/v6.10.2/lib/node_modules/zapier-platform-cli/node_modules/resolve/lib/async.js:105:17)
    at /Users/aymericbouzy/.nvm/versions/node/v6.10.2/lib/node_modules/zapier-platform-cli/node_modules/resolve/lib/async.js:26:73
    at FSReqWrap.oncomplete (fs.js:123:15)
```

Previously it was working fine ...

BrunoBernardino:
Fixes #189 

@FokkeZB Tagging you in case you want to test with your app (I've tested and it worked).

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "archiver": "1.3.0",
     "async": "2.1.5",
     "babel-polyfill": "6.23.0",
-    "browserify": "14.4.0",
+    "browserify": "16.2.0",
     "buffer": "5.0.6",
     "cli-table2": "0.2.0",
     "colors": "1.1.2",
--- a/src/utils/build.js
+++ b/src/utils/build.js
@@ -65,7 +65,7 @@ const requiredFiles = (cwd, entryPoints) => {
       'Buffer.isBuffer': undefined,
       Buffer: undefined
     },
-    ignoreMissing: false,
+    ignoreMissing: true,
     debug: false,
     standalone: undefined
   };
@@ -74,18 +74,11 @@ const requiredFiles = (cwd, entryPoints) => {
   return new Promise((resolve, reject) => {
     b.on('error', reject);
 
-    const ignoredDependencies = [
-      'debug' // Causes https://github.com/zapier/zapier-platform-cli/issues/189
-    ];
-
-    b.exclude(ignoredDependencies);
-
     const paths = [];
     b.pipeline.get('deps').push(
       through
         .obj((row, enc, next) => {
           const filePath = row.file || row.id;
-          // why does browserify add /private + filePath?
           paths.push(stripPath(cwd, filePath));
           next();
         })
--- a/src/utils/build.js
+++ b/src/utils/build.js
@@ -74,6 +74,12 @@ const requiredFiles = (cwd, entryPoints) => {
   return new Promise((resolve, reject) => {
     b.on('error', reject);
 
+    const ignoredDependencies = [
+      'debug' // Causes https://github.com/zapier/zapier-platform-cli/issues/189
+    ];
+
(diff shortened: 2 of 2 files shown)


BrunoBernardino:
Could it be because you didn't `npm install --save axios` (so it wasn't added to the package.json)?

It could break if it's _really_ required (usually isn't), but to include it, it should be possible with `--disable-dependecy-detection`.

eliangcs:
> It could break if it's really required (usually isn't)

@BrunoBernardino not sure if I'm missing something. But in the case of `axios`:

```
├─┬ axios@0.17.1 
│ ├─┬ follow-redirects@1.2.6 
│ │ └─┬ debug@3.1.0 
```

`follow-redirects` requires `debug` in its [`index.js`](https://github.com/olalonde/follow-redirects/blob/1b6340f83ad5596a0a38c16a7113692bd90301f2/index.js#L6). So even if `zapier build` succeeds to generate a zip package, it looks like if we don't include `debug` in the built zip package, the `follow-redirects` could break when the app gets run on the our server.

In #189, `zapier build` fails because:

1. `npm install` doesn't install `support-colors` because `support-colors` is not a hard requirement, i.e., it's required by `debug` during runtime
2. When `browserify` tries to find all the dependencies for us, it thinks `support-colors` is a hard requirement and tries to pull it, but can't find it in `node_moduls`

To fix it, I think we should do something smarter in step 2: say when `browserify` can't find a module (e.g. `support-colors`), we do a `npm install support-colors` under the hood.

BrunoBernardino:
Good point. That won't be easy, though.

What if we just ignore dependencies that weren't found (it's up to the dev to install)?

eliangcs:
> Basically, instead of excluding debug or throwing an error on not finding supports-color, do neither, and just keep going.

Agreed, but maybe we should print some warning messages. I guess we can safely assume a dependency is optional if `browserify` can't find it after `npm install` does the job.

BrunoBernardino:
Ok. I'll look into how that's possible later today.

BrunoBernardino:
@eliangcs I looked around but couldn't find an easy way to get a list of the missing dependencies to print a warning on them.

At least now we won't exclude something that _might_ be really required by dependencies.

PR #303, opened 2018-04-19
[WIP] Clarify the meaning of bundle.inputData and bundle.authData in connection label functions

jvaz11:
To Do: 

- [x] elaborate on what bundle.inputData means in a Digest Auth connectionLabel 
- [ ] add a note to the [Bundle Object](https://zapier.github.io/zapier-platform-cli/#bundle-object) section to clarify that `bundle.inputData` doesn't necessarily represent user-defined data in certain contexts (custom auth, possibly more)
- [ ] add an FAQ that explains Connection Labels in more detail as we don't discuss it outside of the code snippets. perhaps borrow from: https://zapier.com/developer/documentation/v2/appdevguide-auth/#connection-label

### DIFF
--- a/snippets/digest-auth.js
+++ b/snippets/digest-auth.js
@@ -5,8 +5,9 @@ const authentication = {
     url: 'https://example.com/api/accounts/me.json'
   },
   connectionLabel: (z, bundle) => {
-    // Can also be a string, check basic auth above for an example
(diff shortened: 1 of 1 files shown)


xavdid:
@jvaz11 is this still ongoing?

jvaz11:
Sorry about the noise here—closing this out as I'm not 100% sure where I was going with this back in April 😄 .

PR #309, opened 2018-05-01
Add cursor docs

xavdid:
This will go out with https://github.com/zapier/zapier-platform-core/pull/76. 

As with all docs, the PR is overly messy. The only important pieces are the snippet files and the two blurbs in `README-source`

Jacob, requested you since you were in a ticket where someone was complaining about how confusing the original paging snippet was. Also with docs, it'a always nice to get fresh eyes to see if you think you could come in and use this feature with only these to go on.

Thanks a bunch!


README-source.md
```diff
@@ -1339,8 +1348,8 @@ Paging is **only used when a trigger is part of a dynamic dropdown**. Depending
 
 Paging is a lot like a regular trigger except the range of items returned is dynamic. The most common example of this is when you can pass a `start` parameter:
```
codekirei: "when you can pass a `start` parameter:" 
-> 
"when you can pass an `offset` parameter:"
?

I know this isn't something you changed in this PR, but it's odd we say `start` and then use `offset` just below in the example

snippets/paging-cursor.js
```diff
@@ -0,0 +1,48 @@
...
+          params: { cursor: cursor } // if cursor is null, it's ignored here
+        }
+      );
+    })
+    .then(response => {
+      // need to save the cursor and return a promise, but also need to pass the data along
+      return Promise.all([response.items, z.cursor.set(response.nextPage)]);
+    })
+    .then(promises => {
+      // [items[], null]
```
codekirei: I'd go ahead and use destructuring here unless we're intentionally not doing that to try to keep the docs really simple

```js
.then(([items, /* null */]) => {
  return items;
});
```

snippets/paging-cursor.js
```diff
@@ -0,0 +1,48 @@
...
+      return promises[0];
+    });
+};
+
+// ---------------------------------------------------
+
+const performWithAsync = async (z, bundle) => {
+  const cursor = await z.cursor.get(); // string | null
```
codekirei: we probably want to do something similar to `performWithoutAsync`:

```js
let cursor;
if (bundle.meta.page) {
  cursor = await z.cursor.get();
}
```

snippets/paging-cursor.js
```diff
@@ -0,0 +1,48 @@
+// the perform method of our trigger
+// ensure operation.canPaginate is true!
+
+const performWihoutAsync = (z, bundle) => {
```
codekirei: performWi**t**houtAsync

README-source.md
```diff
@@ -1352,7 +1361,15 @@ const getList = (z, bundle) => {
...
 ```
 
-Lastly, you need to set `canPaginate` to `true` in your polling definition (per the [schema](https://github.com/zapier/zapier-platform-schema/blob/master/docs/build/schema.md#basicpollingoperationschema)).
+If your API uses cursor-based paging instead of an offset, you can use `z.cursor.get` and `z.cursor.set`:
+
+```js
+[insert-file:./snippets/paging-cursor.js]
+```
+
+Cursors are stored per-zap and last about an hour. Per the above, make sure to only include the cursor if `bundle.meta.page === 0`, so you don't accidentally reuse a cursor from a previous poll.
```
codekirei: Don't we only want to use the cursor if `bundle.meta.page !== 0`?

snippets/async-polling.js
```diff
@@ -1,39 +1,39 @@
...
+    params: {
+      _start: start,
+      _limit: limit
     }
+  });
+
+  let results = response.json;
+
+  // keep paging until the last item was created over two hours ago
+  // then we know we almost certainly havne't missed anything and can let
```
codekirei: hav**en**'t

snippets/async-polling.js
```diff
@@ -1,39 +1,39 @@
...
+  //   deduper handle the rest
+
+  while (new Date(results[results.length - 1].createdAt) < hoursAgo) {
```
codekirei: isn't this backwards? posts newer than 2 hours ago will have `createdAt > hoursAgo`, yeah?

also, I'd suggest having a "bail out" counter var in case something goes wrong, like

```js
let cycles = 0;
while (
  new Date(results[results.length - 1].createdAt) > hoursAgo &&
  cycles++ < 10
) {
  // ...
}
```

snippets/paging-cursor.js
```diff
@@ -0,0 +1,48 @@
...
+  return Promise.resolve()
+    .then(() => {
+      if (bundle.meta.page === 0) {
+        // first page, no need to fetch a cursor
+        return Promise.resolve();
+      } else {
```
codekirei: unnecessary `else` — since the `if` has a `return`, you can just `return z.cursor.get()` directly at the top of this `.then()` callback's scope

snippets/paging-cursor.js
```diff
@@ -0,0 +1,48 @@
...
+
+// ---------------------------------------------------
+
+const performWithAsync = async (z, bundle) => {
+  const cursor = await z.cursor.get(); // string | null
+
+  const response = await z.request(
+    'https://5ae7ad3547436a00143e104d.mockapi.io/api/recipes',
+    {
+      params: { cursor: cursor } // if cursor is null, it's ignored here
```
codekirei: > if cursor is null

Same when it's `undefined`? Since that's the more likely scenario (I think?), I'd lean towards stating that directly.

```
> null == undefined
true
> null === undefined
false
```

snippets/async-polling.js
xavdid: good call on the `<`! I'm going to leave off the bailout because it might lead to missed triggers and overcomplicates the docs

snippets/paging-cursor.js
xavdid: fair point. I know that's what we usually do, but in the interest of simplicity, i'm going to leave as-is unless you care a lot

snippets/paging-cursor.js
xavdid: I dug in here - we use node's `querystring` module. non-zero falsy values get passed in without values: 

`_limit=3&_start=3&null=&undef=&zero=0&emptyStr=`

That said, the cursor response explicitly returns `null` when there's no cursor value. I can update that doc to be more clear about what happens to null values in a querystring though, since it could (though i don't think it's supposed to) matter on the external server

snippets/paging-cursor.js
codekirei: fine by me to leave as-is 👍

xavdid:
Woof, that was a weird one. travis was only failing on node 8 because of a missing dep. Not sure what the bug exactly is, but there's an npm bug where some required dependencies aren't in the package-lock.json. The fix: run `npm install` *twice*, commit the new (correct) package-lock.json. ¯\\\_(ツ)_/¯

Maybe we should switch these to yarn.

PR #321, opened 2018-05-18
Build and upload source.zip

BrunoBernardino:
This will allow our tech support and app reviews teams to better troubleshoot and help developers.

While https://github.com/zapier/zapier/pull/17580 is required for us to benefit from this, we can merge this without shipping that, as we'll just be "losing" the `source.zip` until then.


src/commands/upload.js
```diff
@@ -2,12 +2,11 @@ const constants = require('../constants');
 const utils = require('../utils');
 
 const upload = context => {
-  const zipPath = constants.BUILD_PATH;
   context.line('Preparing to upload a new version.\n');
-  return utils.upload(zipPath, '.').then(() => {
+  return utils.upload().then(() => {
```
xavdid: I think this `upload()` call will result in an error. The third param (`appDir`) is used directly in `path.resolve()`, which only accepts strings and throws on `undefined`. Could use a default param (we're using babel, might as well use some fancy features) or give it a `|| ''` [here](https://github.com/zapier/zapier-platform-cli/pull/321/files#diff-93af3604864045d869b58d65af952dedR264).

BrunoBernardino: Oh, nice catch! I only tested with `zapier push` and `zapier build`, not `zapier upload`! 🤦‍♂️

PR #326, opened 2018-06-08
Ask for promote when fully migrating a public app

eliangcs:
Addresses [PDE-160](https://zapierorg.atlassian.net/browse/PDE-160).

It makes little sense to migrate 100% without a promotion beforehand. This PR adjusts the behavior of `zapier migrate` so that when a dev does a 100% migration on a public app, we ask if the dev wants to promote that version first.

As for private apps, the 100% migration should automatically do an implicit promotion for the devs. It's already being implemented in the backend, so we don't need to handle that here.

### DIFF
--- a/src/commands/migrate.js
+++ b/src/commands/migrate.js
@@ -19,7 +19,7 @@ const migrate = (context, fromVersion, toVersion, optionalPercent = '100%') => {
   return utils
     .getLinkedApp()
     .then(app => {
-      let promoteFirst = false;
+      let answerPromise = false;
       if (
         optionalPercent === 100 &&
         app.public &&
@@ -39,15 +39,15 @@ const migrate = (context, fromVersion, toVersion, optionalPercent = '100%') => {
           if (!yes && !no) {
             throw new Error('That answer is not valid. Please try "y" or "n".');
           }
-          return yes;
+          return true;
         };
-        promoteFirst = utils.promiseDoWhile(action, stop);
+        answerPromise = utils.promiseDoWhile(action, stop);
       }
-      return Promise.all([app, promoteFirst]);
+      return Promise.all([app, answerPromise]);
     })
-    .then(([app, promoteFirst]) => {
+    .then(([app, answer]) => {
       let promotePromise = null;
-      if (promoteFirst) {
+      if (hasAccepted(answer)) {
         promotePromise = promote(context, toVersion, false);
       }
       return Promise.all([app, promotePromise]);
--- a/src/commands/migrate.js
+++ b/src/commands/migrate.js
@@ -1,4 +1,11 @@
+const promote = require('./promote');
 const utils = require('../utils');
+
+const hasAccepted = answer =>
+  answer.toLowerCase() === 'y' || answer.toLowerCase() === 'yes';
+
+const hasRejected = answer =>
+  answer.toLowerCase() === 'n' || answer.toLowerCase() === 'no';
 
 const migrate = (context, fromVersion, toVersion, optionalPercent = '100%') => {
   if (!toVersion) {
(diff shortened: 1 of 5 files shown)


eliangcs:
cc @ibolmo for you may be interested in having a look.

src/commands/migrate.js
```diff
@@ -8,9 +15,45 @@ const migrate = (context, fromVersion, toVersion, optionalPercent = '100%') => {
     return Promise.resolve();
   }
   optionalPercent = parseInt(optionalPercent, 10);
+
+  let app;
```
ibolmo: any reason to scope this outside?

src/commands/migrate.js
```diff
@@ -8,9 +15,45 @@ const migrate = (context, fromVersion, toVersion, optionalPercent = '100%') => {
...
+
   return utils
     .getLinkedApp()
-    .then(app => {
+    .then(linkedApp => {
+      app = linkedApp;
+      if (
+        optionalPercent === 100 &&
```
xavdid: is there any value in setting this threshold a little lower? presumably moving 90% of users to a version means you want that version to be public too. I could even see a case for going lower, maybe 75%.

src/commands/migrate.js
```diff
@@ -8,9 +15,45 @@ const migrate = (context, fromVersion, toVersion, optionalPercent = '100%') => {
...
+        const stop = answer => {
+          const yes = hasAccepted(answer);
+          const no = hasRejected(answer);
+          if (!yes && !no) {
+            throw new Error('That answer is not valid. Please try "y" or "n".');
+          }
+          return answer;
+        };
+
+        return utils.promiseDoWhile(action, stop);
```
xavdid: not necessarily something to fix in this PR, but it would be worth abstracting input like this into a `prompUser(question: string, acceptedAnswers?: string[])`-type method. We could also use a library since it seems like we're doing a big of hack here with the promise loop. ¯\\\_(ツ)_/¯

src/commands/promote.js
```diff
@@ -9,7 +9,7 @@ const hasCancelled = answer =>
 const hasAccepted = answer =>
   answer.toLowerCase() === 'y' || answer.toLowerCase() === 'yes';
 
-const promote = (context, version) => {
+const promote = (context, version, printMigrateHint = true) => {
```
xavdid: I like it!

src/commands/migrate.js
eliangcs: Thanks for the suggestion! Fixed in 032f11e with `Promise.all`.

src/commands/migrate.js
eliangcs: It's hard to define such a threshold. You can always ask why this value and why not that value. So I think it's easier for us to assume when a dev does a partial migration, they're testing. In addition, for private apps, the backend does implicit promotion (such as moving ZTs to the new version) only if it's a 100% migration. To make it symmetry on the public apps, shouldn't we only suggest for promotion only when it's 100% migration?

xavdid: Yeah, that's fair! It's a pattern devs have some issues with, but maybe the consistency is worthwhile

src/commands/migrate.js
eliangcs: @xavdid totally agreed. I made it a function named `getYesNoInput` in 7510d13. Can I have another review? Thanks!

eliangcs:
@xavdid `async / await` style is definitely something I'm looking forward to seeing. Coding in promises has been costing a lot of my brain power.

xavdid:
No need to wait! We're already running babel on `cli`, so write something with async (and remember to build) and you're good to go. 

I'm not sure why we haven't been writing stuff that way before. Maybe just habit?

eliangcs:
@xavdid I know we can start using `async / await`. But promises are already everywhere, so I guess we've been writing in promises for consistency until there's a big refactor that moves all of them to `async / await`?

xavdid:
that's fair! Since they work with one another it might be easier to introduce them as we go. i'm excited for that future though

PR #330, opened 2018-06-13
Whitelist files in package

eliangcs:
Whitelist instead of blacklisting files we want to publish: https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d. Thanks for the info @xavdid!


package.json
```diff
@@ -7,6 +7,11 @@
   "author": "Bryan Helmig <bryan@zapier.com>",
   "license": "UNLICENSED",
   "main": "zapier.js",
+  "files": [
+      "lib/*.js",
+      "lib/commands/*.js",
+      "lib/utils/*.js"
```
xavdid: sweet! There's just a couple of things we're missing that i'm 99% sure we need:

* zapier.js
* scaffold/

package.json
```diff
@@ -7,6 +7,11 @@
   "author": "Bryan Helmig <bryan@zapier.com>",
   "license": "UNLICENSED",
   "main": "zapier.js",
+  "files": [
+      "lib/*.js",
```
xavdid: I think this can be simplified to include the whole `lib` directory, maybe excluding bin (`"!lib/bin/`)

package.json
eliangcs: * `zapier.js` is specified as `main`, so npm includes it even if we don't specify it in `files`. But on the second thought, it always better to be explicit, so added in 8a1f125.
* Nice catch on `/scaffold/`. Added in a1e9fb4.

package.json
eliangcs: I feel whitelisting is more straightforward and I'm not sure if we can do exclusion in `package.json`'s `files`. Can't find it in the npm's [doc](https://docs.npmjs.com/files/package.json#files).

eliangcs:
@xavdid updated the PR. Can you re-review? Thanks!

package.json
xavdid: ah, you're right about main. that's safe to ignore, but doesn't hurt to have.

package.json
xavdid: npm uses [Glob](https://github.com/isaacs/node-glob#glob-primer), so we can use anything that's valid in that. I don't hate the simplicity though, so it's your call

xavdid:
awesome, thanks!

PR #350, opened 2018-07-25
Include the right binary for deasync

eliangcs:
We run the apps on Node 8 in AWS Lambda since CLI 7.0.0, but we're still shipping the Node 6 binary for [deasync](https://github.com/abbr/deasync) (dependency of [legacy-scripting-runner](https://github.com/zapier/zapier-platform-legacy-scripting-runner)). This one-liner change should fix it.


eliangcs:
@xavdid totally agreed. Fixed in 08b380d.

PR #361, opened 2018-09-21
Add smoke tests

eliangcs:
"Smoke tests" - these are some sanity tests I've been doing manually every time I release a new version of the npm package. This is to make sure there're no mistakes during packaging, which can't be captured in unit tests. Let's automate these smoke tests so we can be more confident about releasing.

Going to do the same to core and schema as well.


src/smoke-tests/smoke-tests.js
```diff
@@ -0,0 +1,121 @@
...
+      filename: null,
+      version: null,
+      path: null
+    },
+    workdir: null,
+    cliBin: null
+  };
+
+  before(() => {
+    if (process.env.DEPLOY_KEY) {
```
eliangcs: `DEPLOY_KEY` is already configured on Travis and it associates to a test account I created on Zapier.com.

package.json
```diff
@@ -28,6 +28,7 @@
     "lint": "node_modules/.bin/eslint zapier.js src",
     "lint-snippets": "node_modules/.bin/eslint snippets --rule 'no-unused-vars: 0' --ignore-pattern snippets/next.js",
     "test": "npm run build && NODE_ENV=test node_modules/mocha/bin/mocha -t 30000 --recursive lib/tests && npm run lint && npm run lint-snippets",
+    "smoke-test": "npm run build && NODE_ENV=test node_modules/mocha/bin/mocha -t 30000 --recursive lib/smoke-tests",
```
xavdid: `node_modules/mocha/bin/mocha` can just be `mocha` (same with above line)

xavdid: also, might make a `pretest` script that runs build and run it manually from here (it'll get picked up before `test` automatically)

src/smoke-tests/smoke-tests.js
xavdid: cool! is that anywhere in 1P or somewhere others will be able to access it?

src/smoke-tests/smoke-tests.js
```diff
@@ -0,0 +1,121 @@
...
+    const lines = npmPack.stdout.split('\n');
+    for (let i = lines.length - 1; i >= 0; i--) {
+      const line = lines[i].trim();
+      if (line) {
+        context.package.filename = line;
+        break;
+      }
+    }
+    context.package.version = context.package.filename.match(
+      /\d+\.\d+\.\d+/
```
xavdid: might be worth moving this regex into a constant so it can be enforced the same way across the package (since I bet we use the same pattern elsewhere)

src/smoke-tests/smoke-tests.js
```diff
@@ -0,0 +1,121 @@
...
+    let res = await fetch(baseUrl);
+    const packageInfo = await res.json();
+    const latestVersion = packageInfo['dist-tags'].latest;
+
+    res = await fetch(`${baseUrl}/-/zapier-platform-cli-${latestVersion}.tgz`, {
+      method: 'HEAD'
+    });
+    const baselineSize = res.headers.get('content-length');
+    const newSize = fs.statSync(context.package.path).size;
+    newSize.should.be.within(baselineSize * 0.7, baselineSize * 1.3);
```
xavdid: Very cool 😍

src/smoke-tests/smoke-tests.js
eliangcs: Yes! The account is cli-test@zapier.rodeo. The deploy key can be found there in the Form Details.

src/smoke-tests/smoke-tests.js
eliangcs: Made it a constant in 00f2354. I couldn't find anywhere else in the cli repo that uses this regex. So I think we can keep it in this file (`smoke-tests.js`) for now.

package.json
eliangcs: Nice suggestion! Fixed in 00f2354.

PR #368, opened 2018-10-23
Add digest-auth example

eliangcs:
Adds [digest-auth](https://github.com/zapier/zapier-platform-example-app-digest-auth) example app. Relates to https://github.com/zapier/zapier-platform-core/pull/123.

### DIFF
--- a/README-source.md
+++ b/README-source.md
@@ -299,6 +299,20 @@ Useful if your app requires two pieces of information to authentication: `userna
 [insert-file:./snippets/basic-auth.js]
 ```
 
+### Digest
+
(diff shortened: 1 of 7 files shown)


eliangcs:
@xavdid thanks for mentioning! A programmatic list is a good idea. But before we have that, I just added digest-auth to the wiki.

Also, I've pushed a doc update dc4792b on digest auth. Would like to know if you have any feedback. Thanks!

xavdid:
Cool! Left a couple of minor things that you can adapt as you feel necessary.

PR #370, opened 2018-10-24
Fix bug where deasync binary isn't in Windows build

### LINKED ISSUE #369: Deasync.node is not included in the zip file when we build from windows os
zapier build and zapier push do not include the deasync.node which is needed from the zapier-platform-legacy-scripting-runner in the build.zip file when we run the build command from windows.
The reason is that the path separator is not '/' in windows so the condition in build.js
```
filePath.endsWith(`/bin/linux-x64-node-${nodeMajorVersion}/deasync.node`)
```
never returns true.

Maybe you could check the separator as you already do in 
```
filePath.match(path.sep === '\\' ? /aws-sdk\\apis\\.*\.json/ : /aws-sdk\/apis\/.*\.json/)
```
 
* CLI Version (`zapier -v`): 7.3.0
* Operating System and version: windows 10

eliangcs:
Fixes #369.

### DIFF
--- a/src/utils/build.js
+++ b/src/utils/build.js
@@ -180,9 +180,7 @@ const forceIncludeDumbPath = (appConfig, filePath) => {
     filePath.endsWith('definition.json') ||
     filePath.endsWith(
       // Special, for zapier-platform-legacy-scripting-runner
-      path.sep === '\\'
-        ? `\\bin\\linux-x64-node-${nodeMajorVersion}\\deasync.node`
-        : `/bin/linux-x64-node-${nodeMajorVersion}/deasync.node`
+      path.join('bin', `linux-x64-node-${nodeMajorVersion}`, 'deasync.node')
     ) ||
     filePath.match(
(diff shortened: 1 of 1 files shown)


src/utils/build.js
```diff
@@ -178,7 +178,12 @@ const forceIncludeDumbPath = (appConfig, filePath) => {
   return (
     filePath.endsWith('package.json') ||
     filePath.endsWith('definition.json') ||
-    filePath.endsWith(`/bin/linux-x64-node-${nodeMajorVersion}/deasync.node`) || // Special, for zapier-platform-legacy-scripting-runner
+    filePath.endsWith(
+      // Special, for zapier-platform-legacy-scripting-runner
+      path.sep === '\\'
```
xavdid: You can change this whole ternary to use `path.join`, which takes care of detecting and using the separator correctly. Same goes for above, and basically any other use of `path.sep`.

```js
filePath.endsWith(path.join('bin', `linux-x64-node-${nodeMajorVersion}`, 'deasync.node'))
```

PR #372, opened 2018-10-24
Adjust convert scaffold to handle test triggers that return arrays.

stevemolitor:
WIP - not fully tested yet!

Web builder auth test triggers can return arrays, which the CLI does not expect. So, shim it such that if the test trigger returns an array, return the first element of the array.

### DIFF
--- a/scaffold/convert/session.template.js
+++ b/scaffold/convert/session.template.js
@@ -40,7 +40,7 @@ const getConnectionLabel = (z, bundle) => {
 const authentication = {
   // TODO: just an example stub - you'll need to complete
   type: 'session',
-  test: testTrigger.operation.perform,
+  test: performAuthTest(testTrigger.operation.perform),
   fields: [
 <%= FIELDS %>
   ],
--- a/scaffold/convert/session.template.js
+++ b/scaffold/convert/session.template.js
@@ -1,3 +1,4 @@
+const { ensureObject } = require('./utils');
 const testTrigger = require('<%= TEST_TRIGGER_MODULE %>');
 
 const getSessionKey = (z, bundle) => {
@@ -40,7 +41,7 @@ const getConnectionLabel = (z, bundle) => {
 const authentication = {
   // TODO: just an example stub - you'll need to complete
   type: 'session',
-  test: performAuthTest(testTrigger.operation.perform),
+  test: ensureObject(testTrigger.operation.perform),
   fields: [
 <%= FIELDS %>
   ],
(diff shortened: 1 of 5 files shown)


stevemolitor:
I haven't tested this yet all the way thru by deploying an app and running a zap, but I picked this up on the end of my support shift so I thought I'd put it out there as is for feedback.

scaffold/convert/utils.template.js
```diff
@@ -30,6 +30,19 @@ const runBeforeMiddlewares = (request, z, bundle) => {
...
+// does not handle. So, if an array is returned from the test
+// trigger, grab the first result and return that. 
+const performAuthTest = testTrigger =>
+  testTrigger.operation.perform(z, bundle).then(response => {
+    if (Array.isArray(response)) {
+      // if legacy test trigger returns an array, return the first element:
+      return response[0];
+    }
+    return response;
+  });
```
eliangcs: `z` and `bundle` aren't defined here. The function signature should be the same as the wrapped function (i.e. `testTrigger`). I also think this could be named something more generic, such as `ensureObject`. So I'd write:

```js
const ensureObject = perform => {
  return (z, bundle) => perform(z, bundle).then(result => {
    if (Array.isArray(result) && result.length > 0) {
      return response[0];
    }
    return response;
  });
};
```

scaffold/convert/session.template.js
```diff
@@ -40,7 +40,7 @@ const getConnectionLabel = (z, bundle) => {
 const authentication = {
   // TODO: just an example stub - you'll need to complete
   type: 'session',
-  test: testTrigger.operation.perform,
+  test: performAuthTest(testTrigger.operation.perform),
```
eliangcs: `performAuthTest` should be imported at the top of this file.

scaffold/convert/simple-auth.template.js
```diff
@@ -17,7 +17,7 @@ const authentication = {
   // TODO: just an example stub - you'll need to complete
   type: '<%= TYPE %>',
 <% if (TEST_TRIGGER_MODULE) { %>
-  test: testTrigger.operation.perform,
+  test: performAuthTest(testTrigger.operation.perform),
```
eliangcs: Same. `performAuthTest` should be imported at the top of this file.

stevemolitor:
Thanks for the great review @eliangcs! Sorry for the slow progress on this one. I implemented your suggested changes here: 4e9fc0700a58b94a1611924bdb55c9fa9ab9e271.

I'm still having a little trouble properly testing this.

eliangcs:
Tested and confirmed it works. Thanks for making those changes, @stevemolitor! I'm going to go ahead and merge this. I hope you don't mind.

eliangcs:
Fix https://github.com/zapier/zapier-platform-cli/pull/372 has been merged and will be shipped in the next CLI release.

PR #373, opened 2018-10-29
Add flowchart to show how a CLI app works (for a polling trigger)

FokkeZB:
As updated and found at https://www.lucidchart.com/documents/edit/df32bce2-14d1-4f91-9621-16665e839642/3

Preview:

https://github.com/zapier/zapier-platform-cli/blob/65ccf4c2cb2a595dac28bc9598d7544cc98b7730/README.md#how-does-the-cli-platform-work

### DIFF
--- a/README-source.md
+++ b/README-source.md
@@ -50,6 +50,12 @@ We then make calls to execute the operations your App defines as we execute Zaps
 Your App takes the input data we provide (if any), makes the necessary HTTP calls,
 and returns the relevant data, which gets fed back into Zapier.
 
+Here's a flow chart for how that works for a simple, polling based trigger:
+
+![](https://cdn.zapier.com/storage/photos/98ee1d46bbe5f5c11077fd83e7aed2f4.png)
+
+> The dashed lines indicate that if you define `perform` as a function instead of a request configuration, you can still use `z.request()`, `throwForStatus`, `z.JSON.parse` and other tools packaged in `zapier-platform-core`.
+
 ### CLI vs the Web Builder Platform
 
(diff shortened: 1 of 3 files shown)


ZaneLyon:
I think this is awesome information to present!  I suspect we need to do more to provide context in order for folks outside Zapier to totally absorb the content.  Just a little bootstrapping of some concepts that we all have ingrained from talking about this stuff all day, every day.

Pretty much every term here needs to be introduced and defined from the perspective of someone outside our org.  What is zapier-platform-core?  What is the build bundle?  Which lanes are running in lambda vs running on the same system that users are interacting with? Explicitly pointing out that the "app scripting" lane is what happens when you use a function in a perform field, where as configuring a request-schema allows the core to handle the request in a standard way based on the meta data.

Something to help overlay what they see - or will see when they get started - on top of this model.  

If we can make this info digestible to the developer new to Zapier, it'll make the rest of the concepts easier to fathom.  

Next: What would be cool is to also show a similar model for what happens when you push a version of a integration.

FokkeZB:
I'll close this as I don't have time to work on this any further.

CC @maguay to inform your work on developer docs.

maguay:
@FokkeZB This is great—and yes, we do need to do more graphics. @aekaplan was looking into being able to make more for our documentations.

Eventually the new platform site will have our new web builder and CLI docs together, along with what I'm hoping is better content that explains Zapier's concepts and helps people understand _what_ they're building and how it works better. And somewhere along the way, sample apps and possibly a restarted platform/development blog (in which shape or form, I'm not sure) would all be things I'd love to see happen. We should definitely talk about this more :)

PR #376, opened 2018-11-02
Rebuild docs

eliangcs:
Looks like the built docs aren't reflecting the latest changes. Let's rebuild the docs.


xavdid:
Odd! They looked like they built before I merged.

PR #380, opened 2018-12-13
Server-side `zapier convert`

eliangcs:
Reworks the `zapier convert` command so that it uses the up-to-date converter implementation on the backend. Requires https://github.com/zapier/zapier/pull/22060.

This PR is quite big and can be unstraightforward to review, but the big ideas are:

* The server now has a `cli-dump` endpoint (see https://github.com/zapier/zapier/pull/22060) that dumps a legacy WB app to a JSON app definition compatible with the CLI platform.
* `zapier convert` pulls the app definition from the `cli-dump` endpoint.
* `zapier convert` splits the app definition, a big JSON blob, into separate js files.
* Most of the conversion logic is moved out of the CLI client to the server.


eliangcs:
@FokkeZB this is something we talked about, so you might be interested in having a look, but please don't feel obligated to review. :)

src/commands/convert.js
```diff
@@ -17,24 +17,27 @@ const convert = (context, appid, location) => {
     return Promise.reject(new Error(message));
   }
 
-  const createApp = tempAppDir => {
-    const url = `${
+  const createApp = async tempAppDir => {
```
xavdid: +1 to eventually getting all of cli to `async // await`!

src/utils/command-invocations.js
```diff
@@ -23,19 +21,27 @@ const maybeLogin = context => {
...
-const maybeRegisterApp = context => {
-  if (!fileExistsSync(constants.CURRENT_APP_FILE)) {
+const maybeRegisterApp = async context => {
+  if (!await hasRegisteredApp()) {
     context.line(
       "Looks like this is your first push. Let's register your app on Zapier."
     );
-    return getInput('Enter app title (Ctrl-C to cancel):\n\n  ').then(title =>
-      register(context, title, { printWhenDone: false })
+    return await getInput('Enter app title (Ctrl-C to cancel):\n\n  ').then(
```
xavdid: minor, seems weird to return the promise chain here. Why not

```js
const input = await getInput('Enter' // ...
await register(context, // ...
return null
```

Also, i'm not sure we need the return statement at all.

src/utils/convert.js
```diff
@@ -1,798 +1,164 @@
...
+};
+
+const prettifyJs = code => prettier.format(code, { singleQuote: true });
+
+const renderTemplate = async (
+  templateFile,
+  templateContext,
+  prettify = true
+) => {
+  const templateBuf = await readFile(templateFile);
```
xavdid: I believe this can be `const template = await readFile(templateFile, 'utf-8')`, since passing encoding gives you a string instead of a buffer.

src/utils/convert.js
```diff
@@ -825,278 +194,257 @@ const renderDefaultInputData = definition => {
...
-  return getHeader(legacyApp).then(header => {
-    templateContext.HEADER = header;
+  return prettifyJs(functionBlock + exportBlock);
+};
 
-    const importLines = [];
+const renderIndex = async appDefinition => {
+  let exportBlock = _.cloneDeep(appDefinition),
+    functionBlock = '',
+    importBlock = '';
```
xavdid: I'd probably do importBlock as an array and then `.join('\n')` at the end instead of needing to use a newline whenever you add to it.

src/utils/convert.js
```diff
@@ -825,278 +194,257 @@ const renderDefaultInputData = definition => {
...
+  endSpinner();
 };
 
-const convertApp = (legacyApp, newAppDir) => {
+const writeZapierAppRc = async newAppDir => {
+  const content = JSON.stringify({
+    includeInBuild: ['scripting.js']
+  });
+  await createFile(content, '.zapierapprc', newAppDir);
+  startSpinner('Writing .zapierapprc');
```
xavdid: presumably we'd want to start the spinner before beginning the async action?

src/utils/convert.js
```diff
@@ -1,798 +1,164 @@
...
-};
-
-const hasAuth = definition => {
-  return (
-    getAuthType(definition) !== 'custom' && !_.isEmpty(definition.auth_fields)
+const createFile = async (content, filename, dir) => {
+  const destFile = path.join(dir, filename);
+  await ensureDir(path.dirname(destFile));
+  await writeFile(destFile, content);
+  startSpinner(`Writing ${filename}`);
```
xavdid: order here is weird too

src/utils/convert.js
```diff
@@ -825,278 +194,257 @@ const renderDefaultInputData = definition => {
...
+};
+
+const convertApp = async (
+  legacyApp,
+  appDefinition,
+  newAppDir,
+  silent = false
+) => {
+  if (silent) {
+    startSpinner = endSpinner = () => null;
```
xavdid: I did some googling and i don't think this has any chance of messing up the import for other files (it's just redefining a local variable). 

That said, I think i'd rather build this functionality into the spinner itself rather than do this each place we may need to maybe not have the spinner actually go. Maybe in utils, instead of exporting the object itself we export a wrapper class?

```js
// wrapper pseudocode
class SpinnerWrapperClass {
  constructor(silent=false)

  startSpinner() {
    if (this.silent) {return}
    spinner.spin()
}

// usage
const {spinner} = require('./utils/display')

const s = new Spinner(silent)
s.startSpinner()
```

slightly more verbose, but then we keep all functionality in the same spot. Not a blocker, just a thought.

src/utils/command-invocations.js
eliangcs: I totally missed it! Cleaned up in 4830272.

src/utils/convert.js
eliangcs: This `readFile` is a function we wrote in `src/utils/files.js` and doesn't accept an encoding option. It is used in other places, so I rather not change its interface for now.

src/utils/convert.js
eliangcs: We're only using it to print messages and not really taking advantage of spinner here. Each output file is generated [concurrently with Promise.all](https://github.com/zapier/zapier-platform-cli/blob/4830272f39329485fc6a00ecb26aa3f691e6308b/src/utils/convert.js#L445). To properly use spinner to show progress, we need to create a spinner instance for each promise, which is not how we do it in `src/utils/display.js`. Since conversion is very fast, spinner doesn't help that much, I wouldn't bother to create multiple spinner instances just to properly show progress.

src/utils/convert.js
eliangcs: See my previous [comment](https://github.com/zapier/zapier-platform-cli/pull/380#discussion_r242788799).

src/utils/convert.js
eliangcs: Currently, the `silent` option is there only for this [test](https://github.com/zapier/zapier-platform-cli/blob/4830272f39329485fc6a00ecb26aa3f691e6308b/src/tests/utils/convert.js#L62) to avoid printing stuff in the console while running tests. It's true that the spinner utils could use some improvement but it's probably not worth doing only for a test.

src/utils/convert.js
eliangcs: Good call! Changed in 3447bd7.

PR #382, opened 2018-12-20
use async // await in build

xavdid:
I originally started this as an improvement to `source.zip` generation (namely, always excluding node_modules) but got distracted and inspired by your PR yesterday that I wanted to do the same. 

No functionality should have changed - this is purely a rewrite using async and await.

### DIFF
--- a/src/utils/build.js
+++ b/src/utils/build.js
@@ -125,6 +125,7 @@ const respectGitIgnore = (dir, paths) => {
     const gitIgnorePath = path.join(dir, '.gitignore');
     const gitIgnoredPaths = gitIgnore(gitIgnorePath);
     const gitFilter = ignore().add(gitIgnoredPaths);
+
     resolve(gitFilter.filter(paths));
   });
 };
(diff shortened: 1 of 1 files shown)


src/utils/build.js
```diff
@@ -321,171 +315,158 @@ const build = (zipPath, sourceZipPath, wdir) => {
...
+      'zapierwrapper.js'
+    )
+  );
+  await writeFile(`${tmpDir}/zapierwrapper.js`, zapierWrapperBuf.toString());
+  endSpinner();
+
+  startSpinner('Building app definition.json');
+  const rawDefinition = (await _appCommandZapierWrapper(tmpDir, {
+    command: 'definition'
+  })).results;
```
eliangcs: I think we need to remove the `.results` part, i.e.,
```js
  const rawDefinition = await _appCommandZapierWrapper(tmpDir, {
    command: 'definition'
  });
```
Otherwise, it will error when you do `rawDefinition.results` again.

src/utils/build.js
```diff
@@ -321,171 +315,158 @@ const build = (zipPath, sourceZipPath, wdir) => {
...
-      // https://blog.pivotal.io/labs/labs/barriers-deterministic-reproducible-zip-files
-      // https://reproducible-builds.org/tools/ or strip-nondeterminism
-      endSpinner();
-      startSpinner('Testing build');
-
-      if (isWindows()) {
-        return {}; // TODO err, what should we do on windows?
+  const fileWriteError = await writeFile(
+    `${tmpDir}/definition.json`,
+    prettyJSONstringify(rawDefinition.results)
```
eliangcs: I think you meant to write

```suggestion
    prettyJSONstringify(rawDefinition)
```

because `.results` has been done previously.

src/utils/build.js
```diff
@@ -286,16 +289,7 @@ const _appCommandZapierWrapper = (dir, event) => {
...
-  wdir = wdir || process.cwd();
-  zipPath = zipPath || constants.BUILD_PATH;
-  sourceZipPath = sourceZipPath || constants.SOURCE_PATH;
-  const osTmpDir = fse.realpathSync(os.tmpdir());
-  const tmpDir = path.join(
-    osTmpDir,
-    'zapier-' + crypto.randomBytes(4).toString('hex')
-  );
-
+const maybeNotifyAboutOutdated = () => {
```
eliangcs: 👍 for making it a function!

src/utils/build.js
```diff
@@ -321,171 +315,158 @@ const build = (zipPath, sourceZipPath, wdir) => {
...
-        }
-        throw new Error(
-          `Unable to write ${tmpDir}/definition.json, please check file permissions!`
-        );
-      }
+const build = async (zipPath, sourceZipPath, wdir) => {
+  wdir = wdir || process.cwd();
+  zipPath = zipPath || constants.BUILD_PATH;
+  sourceZipPath = sourceZipPath || constants.SOURCE_PATH;
+  const osTmpDir = await fse.realpath(os.tmpdir());
```
eliangcs:
```suggestion
  const osTmpDir = fse.realpathSync(os.tmpdir());
```

This needs to be sync version as `fs.realpath` is a [callback-style](https://nodejs.org/docs/latest-v8.x/api/fs.html#fs_fs_realpath_path_options_callback) function.

xavdid: I think this one's ok, since it's using `fs-extra`, which makes everything into promises.

xavdid:
I think we're good on the `fse` front. Updated the rest in a41baab.

src/utils/build.js
eliangcs: 👍 TIL!

PR #388, opened 2019-01-11
Fix bug: `zapier push` doesn't stop on validation errors

### LINKED ISSUE #387: `zapier push` will run `zapier upload` even if `zapier build` fails

While investigating the report at https://zapier-platform.slack.com/archives/C2S5DGF6G/p1547142073081000, I came to the realization that we must have reached this line: https://github.com/zapier/zapier-platform-cli/blob/v7.6.0/src/utils/build.js#L399

But we did not hit this line: https://github.com/zapier/zapier-platform-cli/blob/v7.6.0/src/utils/build.js#L434

The next line logged was `✔ Uploading version 1.0.1`, which occurs in the `upload` function. 

I think? this is caused by not validating the output of `build` at https://github.com/zapier/zapier-platform-cli/blob/v7.6.0/src/utils/build.js#L468, before calling `validate`, but I'm not 100% confident in my knowledge of this code. 

## Your Environment

* CLI Version (`zapier -v`): 7.6.0

eliangcs:
Fixes #387, a bug introduced to 7.6.0 by PR #382.

Steps to reproduce:

1. `zapier init example-app`
2. `zapier push` with some random name
3. create a trigger with a very short description (length < 12)
4. `zapier validate` would show an error
5. bump version to 1.0.1 in `package.json`
6. `zapier push` shows success, but 1.0.1 wasn't really pushed

### DIFF
--- a/src/utils/build.js
+++ b/src/utils/build.js
@@ -405,12 +405,17 @@ const build = async (zipPath, sourceZipPath, wdir) => {
 
   const validationErrors = validateResponse.results;
   if (validationErrors.length) {
-    return {
-      validationErrors: {
-        validation: validationErrors
-      }
-    };
+    if (global.argOpts.debug) {
+      console.log('\nErrors:');
+      console.log(validationErrors);
+      console.log('');
+    }
+
+    throw new Error(
+      'We hit some validation errors, try running `zapier validate` to see them!'
+    );
   }
+
   // No need to mention specifically we're validating style checks as that's
   //   implied from `zapier validate`, though it happens as a separate process
   const styleChecksResponse = await callAPI('/style-check', {


kgodtfredsen:
Heya, how can I tell if this fix has been deployed? I had a customer run into this today and could recreate, so we both still have the older versions installed.

xavdid:
@kgodtfredsen this is not yet deployed. We release CLI every 2 weeks typically (though that schedule was disrupted by holidays and retreat). You can see what's been released in the [changelog](https://github.com/zapier/zapier-platform-cli/blob/master/CHANGELOG.md)

 If anything, I think this is worth a bugfix release sooner rather than later @eliangcs.

xavdid:
@kgodtfredsen `v7.6.1` went out early this morning with the fix in tow! Thanks for reminding us.

kgodtfredsen:
@xavdid thanks much for this work!

PR #390, opened 2019-01-17
update the AWS Lambda supported Node.js links

calebpeterson:
Was reading the docs and encountered what appeared to be a stale link to the AWS Lambda docs for supported Node.js versions.


calebpeterson:
thanks for the quick review and additional pointer @xavdid! built and merged ✔️

PR #393, opened 2019-01-25
update docs to use new meta

xavdid:
Updates the docs to reflect the new `bundle.meta` keys. This won't be relevant until `8.0.0`, so we should be sure to merge this before that goes out

I also tried to specify when each key would be useful. I think that still needs a little tweaking, but this is the idea.

Could also add a link to the [wiki page](https://github.com/zapier/zapier-platform-cli/wiki/bundle.meta-changes) I made for migrations (that we should include in the release notes)


xavdid:
Sure, makes sense. The reason I didn't include it initially was that IMO the docs should be about current content in order to not confuse users. Maybe a link below the table that points to the wiki? So people can find it more easily, but all that's front and center is the good stuff. 

Alternatively, we could put the conversion table in for now and remove it in a couple of versions? I"m not sure. hopefully folks are reading upgrade notes when doing a major bump

xavdid:
Sweet! We can't merge until `8.0.0` is ready to ship, but i'll be ready to go.

PR #403, opened 2019-02-25
[PE-1032] update collaborator to admin

ibolmo:
This updates any mention of collaborator to admin. This follows the new public terminology.


ibolmo:
We'll leave it as is since we couldn't find a good verb equivalent. :D

PR #407, opened 2019-03-13
add login arguments to skip interactivity for better CI/CD integration

mkotlikov:


mkotlikov:
@bryanhelmig please add this feature! Adding this to my CI/CD pipeline is a pain!

bryanhelmig:
We'll take a look @mkotlikov! 🙇

xavdid:
@mkotlikov thanks for the PR! 

It sounds like you're hitting a bug though - deploy keys don't expire unless you call `zapier logout`. Are you seeing different behavior?

LassoMike:
yes, the keys are expiring within 24 hours

xavdid:
Huh. When you have a chance, can you write into `partners@zapier.com` with information about how you (and your team?) are using the CLI? Nothing on our end expires the keys, so it sounds like `zapier logout` is getting run somewhere by someone. I've had the same deploy key on my machine for over a year now and haven't had an issue.

We're reticent to merge this PR because credentials really shouldn't be passed in plaintext. While there's nothing stopping you from using a fork, it's not something we want to encourage most users to do.

LassoMike:
Email to `partners@zapier.com` sent. An option for passing the password like that is common with other CLI tools, as a customer I like having convenience options even if they are less secure, but that's your call! At the same time if a deployment key isn't supposed to expire than I guess this doesn't matter anyways!

xavdid:
this has been resolved - the `zapier logout` command is confusing and should be avoided.

Issue #351, opened 2018-07-26
`zapier upload` should give a meaningful error message when dev hasn't built


eliangcs:
Before a developer builds their project with `zapier build`, running `zapier upload` will just fail silently without any meaningful error messages.

![](https://cdn.zapier.com/storage/photos/5627b065d50eeeb49d6f1c95ba8c7076.png)

## What would you like to change?

If `./build/build.zip` doesn't exist, `zapier upload` should give a hint that they should run `zapier build` first or use `zapier push` command instead.

## Your Environment

* CLI Version (`zapier -v`): 7.1.0
* Operating System and version: Irrelevant

shaunc869:
Here here to that, ran into the same issue!

xavdid:
fixed in #408

PR #408, opened 2019-03-13
Add a better error message when uploading with a zip

stevelikesmusic:
## Description
When a dev attempts to `zapier upload` without having built the integration, we silently fail when trying to read the zip. 

Instead, we'll check to see if the zip exists, and then suggest running `zapier push` if we still need to build.

## Issue
#351

### DIFF
--- a/src/commands/upload.js
+++ b/src/commands/upload.js
@@ -1,15 +1,20 @@
+const colors = require('colors/safe');
 const constants = require('../constants');
 const utils = require('../utils');
 
-const upload = context => {
-  context.line('Preparing to upload a new version.\n');
-  return utils.upload().then(() => {
+const upload = async context => {
+  try {
+    context.line('Preparing to upload a new version.\n');
+    await utils.upload();
+
     context.line(
       `\nUpload of ${constants.BUILD_PATH} and ${
         constants.SOURCE_PATH
       } complete! Try \`zapier versions\` now!`
     );
-  });
+  } catch (error) {
+    context.line(`${colors.red('Problem uploading:')} ${error.message}`);
+  }
 };
 upload.argsSpec = [];
 upload.argOptsSpec = {};
--- a/src/entry.js
+++ b/src/entry.js
@@ -123,7 +123,7 @@ module.exports = argv => {
     process.exit(1);
   }
 
-  commandFunc.apply(commands, [context].concat(args)).catch(err => {
+  commandFunc(context, ...args).catch(err => {
     utils.endSpinner(false);
 
     if (DEBUG || global.argOpts.debug) {
--- a/src/utils/api.js
+++ b/src/utils/api.js
@@ -270,6 +270,13 @@ const upload = (zipPath, sourceZipPath, appDir) => {
 
   const fullZipPath = path.resolve(appDir, zipPath);
   const fullSourceZipPath = path.resolve(appDir, sourceZipPath);
+  const isMissingZip = !fs.existsSync(fullZipPath);
+
+  if (isMissingZip) {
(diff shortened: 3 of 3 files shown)


src/utils/api.js
```diff
@@ -270,6 +270,13 @@ const upload = (zipPath, sourceZipPath, appDir) => {
 
   const fullZipPath = path.resolve(appDir, zipPath);
   const fullSourceZipPath = path.resolve(appDir, sourceZipPath);
+  const isMissingZip = !fs.existsSync(fullZipPath);
+
+  if (isMissingZip) {
+    return Promise.reject(
+      new Error('Missing a built app. Trying running `zapier push`.')
```
stevelikesmusic: Alternatively, we could just `console.error` or `catch` this one level up and `context.line`. I'm in favor of more general error handling in `entry` where this ends up.

xavdid: I think I like that a little more, but it's not a blocker for me

src/commands/upload.js
```diff
@@ -3,13 +3,16 @@ const utils = require('../utils');
...
 const upload = context => {
   context.line('Preparing to upload a new version.\n');
-  return utils.upload().then(() => {
-    context.line(
-      `\nUpload of ${constants.BUILD_PATH} and ${
-        constants.SOURCE_PATH
-      } complete! Try \`zapier versions\` now!`
-    );
-  });
+  return utils
```
xavdid: while you're in here, do you mind converting this to `async`? we've been doing that with other commands as we go, which will eventually remove our dependence on babel.

stevelikesmusic: I can definitely do that! Curious though—what's preventing us from dropping babel now? If it's just waiting to convert to `async/await`, I would think we could run the CLI in node natively.

src/utils/api.js
stevelikesmusic: You're more in favor of `console.error` right here?

src/entry.js
```diff
@@ -123,7 +123,7 @@ module.exports = argv => {
     process.exit(1);
   }
 
-  commandFunc.apply(commands, [context].concat(args)).catch(err => {
+  commandFunc(context, ...args).catch(err => {
```
stevelikesmusic: Left this `.catch` so we didn't grab any unintended `Error`s here.

src/utils/api.js
```diff
@@ -270,6 +270,11 @@ const upload = (zipPath, sourceZipPath, appDir) => {
 
   const fullZipPath = path.resolve(appDir, zipPath);
   const fullSourceZipPath = path.resolve(appDir, sourceZipPath);
+  const isMissingZip = !fs.existsSync(fullZipPath);
+
+  if (isMissingZip) {
+    throw new Error('Missing a built app. Trying running `zapier push`.');
```
xavdid: Thinking about it, this is a little bit of an odd error message. I'd probably change it to "Try running `zapier build`", since push will have automatically run build and if they're seeing this message, it's because they're manually calling `upload` before `build`.

Also there's a typo: `Trying running`

src/entry.js
xavdid: i was worried about this change but it looks like it didn't need to be passing the command object before, so we're good!

src/utils/api.js
stevelikesmusic: Typos... 

My thought is if they're running `upload` without `build`, there's a disconnect with CLI commands. What if we at least say something like `You need to run zapier build first. Or you could run zapier push, which will build and upload in one command.`

xavdid: Love it!

PR #422, opened 2019-04-25
logout by only killing one key, not all

xavdid:
By default, `zapier logout` deactivated all keys. That's odd behavior, so we're doing to make it only kill the active key. This relies on some server work that's not done yet, so there's not much to see yet. Also did some unrelated logging improvements.

### DIFF
--- a/src/commands/logout.js
+++ b/src/commands/logout.js
@@ -3,13 +3,14 @@ const utils = require('../utils');
 
 const logout = async context => {
   context.line(
-    'Preparing to deactivate personal deploy keys and reset local configs.'
+    'Preparing to deactivate local deploy key and reset local configs.'
   );
   context.line();
-  utils.startSpinner('Deactivating personal deploy keys');
+  utils.startSpinner('Deactivating local deploy key');
   try {
     await utils.callAPI('/keys', {
-      method: 'DELETE'
+      method: 'DELETE',
+      body: { single: true }
     });
   } catch (e) {
     // no worries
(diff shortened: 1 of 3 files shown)


src/commands/logout.js
```diff
@@ -3,13 +3,14 @@ const utils = require('../utils');
...
+    'Preparing to deactivate local deploy key and reset local configs.'
   );
   context.line();
-  utils.startSpinner('Deactivating personal deploy keys');
+  utils.startSpinner('Deactivating local deploy key');
   try {
     await utils.callAPI('/keys', {
-      method: 'DELETE'
+      method: 'DELETE',
+      body: { single: true }
```
xavdid: this is the only actual code change

src/utils/api.js
```diff
@@ -85,13 +85,18 @@ const callAPI = (route, options, rawError = false) => {
...
         if (requestOptions.body) {
           const replacementStr = 'raw zip removed in logs';
-          let cleanedBody = _.assign({}, JSON.parse(requestOptions.body), {
-            zip_file: replacementStr,
-            source_zip_file: replacementStr
-          });
+          const requestBody = JSON.parse(requestOptions.body);
+          const cleanedBody = {};
+          for (const k in requestBody) {
+            if (k.includes('zip_file')) {
```
xavdid: this prevents adding a `zip_file` key when there wasn't one before. I was super confused as to why my `DELETE` call had 2 zips attached.

PR #423, opened 2019-04-26
Misc improvements

xavdid:
two small fixes, plus i'm going to fix linting errors in the snippets.


src/utils/build.js
```diff
@@ -121,13 +121,16 @@ const listFiles = dir => {
 
 // Exclude file paths in .gitignore
 const respectGitIgnore = (dir, paths) => {
-  return new Promise(resolve => {
```
xavdid: nothing in here is async, so there was no reason to wrap it all in a promise ¯\\\_(ツ)_/¯

PR #425, opened 2019-04-30
POC for scaffolding an app from a data source

stevelikesmusic:
@xavdid if you want to debug this branch the way it is, the VS Code config is just the default "Launch via NPM":

```js
{
  "type": "node",
  "request": "launch",
  "name": "Launch via NPM",
  "runtimeExecutable": "npm",
  "runtimeArgs": ["run-script", "debug"],
  "port": 9229
},
```


xavdid:
closing in favor of #427, since I didn't actually end up using any of this code but don't want to lose it in a squashed commit

PR #427, opened 2019-05-08
Convert visual apps to CLI. 

xavdid:
Standing on the shoulders of other great work, this ended up being pretty straightforward! 

I liked the idea of re-using the command. That, plus adding the `version` option, makes it pretty straightforward. 

There are a couple of different methods based on differences between what we need for visual/legacy, but the similarity made that part really easy.  

I've been testing with the following:

`zapier convert 17741 a --version=1.0.1`, which is a dev-app-school wistia app. I've confirmed that you can convert, push a new version, and make a new zap (auth + trigger) successfully.


src/commands/convert.js
```diff
@@ -2,7 +2,7 @@ const utils = require('../utils');
 const constants = require('../constants');
 
 const convert = (context, appid, location) => {
-  context.line('Welcome to the Zapier Platform! :-D');
+  context.line('Welcome to the Zapier Platform!');
```
stevelikesmusic: Do we even need this message at all?

xavdid: no, not really!

package.json
```diff
@@ -25,10 +25,12 @@
     "prepublish": "npm run build",
     "build": "rm -rf lib && babel src -d lib --copy-files",
     "watch": "rm -rf lib && babel --watch src -d lib --copy-files",
-    "lint": "eslint zapier.js src",
     "lint-snippets": "eslint snippets --rule 'no-unused-vars: 0' --ignore-pattern snippets/next.js --ignore-pattern 'snippets/dynamic-dropdowns-*'",
+    "lint": "eslint zapier.js src",
+    "postlint": "npm run lint-snippets",
```
stevelikesmusic: Nice

src/commands/convert.js
```diff
@@ -16,64 +16,120 @@ const convert = (context, appid, location) => {
...
 
-    utils.endSpinner();
+      utils.endSpinner();
 
-    return await utils.convertApp(legacyApp, appDefinition, tempAppDir);
+      return utils.convertLegacyApp(legacyApp, appDefinition, tempAppDir);
+    }
   };
 
   return utils.initApp(context, location, createApp).then(() => {
```
stevelikesmusic: Why not instead of asking for the project location, we use the same string that we put in `name` for the `package.json`? (kebab-cased app title) I think it'd be rare that a dev would want those names to be different.

src/tests/utils/convert.js
```diff
@@ -70,19 +265,89 @@ describe('convert', () => {
...
+        'test/triggers/movie.js'
+      ].forEach(filename => {
+        const filepath = path.join(tempAppDir, filename);
+        fs.existsSync(filepath).should.be.true(`failed to create ${filename}`);
+      });
+    });
+  });
+
+  describe.only('visual apps apps', () => {
+    it('should create separate files', async () => {
```
stevelikesmusic: I'm wondering if we need tests to verify the output of files 🤔 For instance, to ensure we get dynamic inputs, functions are correctly written, files are ordered correctly, etc. 

Maybe overkill?

src/utils/convert.js
```diff
@@ -408,21 +445,30 @@ const writeGitIgnore = async newAppDir => {
...
+  const content = prettifyJSON({
+    id
+  });
+  await createFile(content, '.zapierapprc', newAppDir);
+  startSpinner('Writing .zapierapprc');
+  endSpinner();
+};
+
+const convertApp = async (appInfo, appDefinition, newAppDir, opts = {}) => {
+  const defaultOpts = { silent: false, legacy: true };
```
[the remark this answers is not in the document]
xavdid: the only reason is the tests. 

with silent: 

```
$ NODE_ENV=test mocha -t 50000 --recursive lib/tests


  convert
    visual apps apps
      ✓ should create separate files (110ms)


  1 passing (125ms)

✨  Done in 3.65s.
```

without silent:

```
$ NODE_ENV=test mocha -t 50000 --recursive lib/tests


  convert
    visual apps apps
✔ Writing .gitignore
✔ Writing triggers/project.js
✔ Writing package.json
✔ Writing creates/create_project.js
✔ Writing .env
✔ Writing authentication.js
✔ Writing .zapierapprc
✔ Writing .zapierapprc
✔ Writing triggers/codemode.js
✔ Writing index.js
✔ Writing test/creates/create_project.js
✔ Writing test/triggers/project.js
✔ Writing test/triggers/codemode.js
      ✓ should create separate files (115ms)


  1 passing (132ms)

✨  Done in 3.66s.
```

Could get a similar effect by only logging if `NODE_ENV !== 'TEST'` if we wanted to simplify it a little bit

src/tests/utils/convert.js
```diff
@@ -41,6 +41,157 @@ const appDefinition = {
...
+            removeMissingValuesFrom: {},
+            headers: {
+              'Content-Type': 'application/json',
+              Authorization: 'Bearer {{bundle.authData.access_token}}',
+              Accept: 'application/json'
+            },
+            params: {},
+            method: 'POST'
+          },
+          inputFields: [
```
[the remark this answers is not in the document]
xavdid: can you clarify what you mean by dynamic input field? I know in CLI we can have custom fields, but I didn't think that was something that was possible in godzilla

src/tests/utils/convert.js
```diff
@@ -41,6 +41,157 @@ const appDefinition = {
...
+        },
+        noun: 'Code',
+        display: {
+          hidden: false,
+          important: true,
+          description: "just runs some code, let's go",
+          label: 'New Code Trigger'
+        },
+        key: 'codemode'
+      }
```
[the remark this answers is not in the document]
xavdid: I think this is actually the preferred way to do it. Since you need to have both the search and create imported, it usually makes more sense to have it in the index. that also matches the example: https://github.com/zapier/zapier-platform-example-app-search-or-create/blob/017bb5ff8c09bfadf980e88ced6d28741e50e7ac/index.js#L27-L39

src/utils/convert.js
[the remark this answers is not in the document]
stevelikesmusic: Thanks @xavdid! That makes sense. 

`NODE_ENV` could be a good option. No strong feelings one way or the other.

package.json
```diff
@@ -25,10 +25,12 @@
...
     "pretest": "npm run build",
-    "test": "NODE_ENV=test mocha -t 50000 --recursive lib/tests && npm run lint && npm run lint-snippets",
+    "test": "NODE_ENV=test mocha -t 50000 --recursive lib/tests",
+    "/posttest": "npm run lint",
```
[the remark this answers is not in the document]
xavdid: I like that way more!

src/tests/utils/convert.js
```diff
@@ -41,6 +41,157 @@ const appDefinition = {
...
 };
 
+const visualAppDefinition = {
+  definition_override: {
+    platformVersion: '8.0.1',
+    creates: {
+      create_project: {
+        operation: {
+          perform: {
+            body: {
```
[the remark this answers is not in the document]
xavdid: I think we do want to keep the shorthand requests. They provide some nice auto error-handling and other sugar that are perfectly reasonable to rely on if you don't need a more complex request. Curlies should also be fine, though #428 makes me a little worried about that.

src/commands/convert.js
xavdid: mostly so that devs can run the command from anywhere. `zapier convert 1234 ../../projects`. Since we're writing (and potentially overwriting) files, seems good to be explicit about it.

src/tests/utils/convert.js
xavdid: nah, more tests are good! I can cover a couple of key cases in there.

xavdid:
Great questions! 

1. I like being explicit about the version. It also reduces the number of API calls, since we only check for WB _or_ godzilla. We could probably make both calls every time because it's unlikely that a dev has access to both CLI app 1234 and WB app 1234. We could also do a nice dropdown selector if we wanted, though that's probably out of scope here (we'd want to switch our prompt library from custom to off-the-shelf which is something I want to do anyway, but maybe not here)
2. ordering could be nice, I'll look at that!
2. I like that idea, but it sounds harder to maintain. As far as I can tell, you're proposing the following:

```js
const convertLegacyApp(...) => {
  a()
  b()
  c()
  d()
}

const convertVisualApp(...) => {
  a()
  b()
  j()
  k()
  d()
}
```

where we move a lot of the logic into the reused functions. I like that, but worry that if we decide we want to add something after `b`, we now have to add that in multiple places (currently 2, but maybe more in the future). IMO, sounds tougher. But let me know if I'm misunderstanding!

src/commands/convert.js
```diff
@@ -16,64 +16,120 @@ const convert = (context, appid, location) => {
...
     context.line(
-      `Finished! You may try \`npm install\` and then \`zapier test\` in "${location}" directory.`
+      `Finished! You may try \`npm install\` and then \`zapier test\` in the "${location}" directory.`
     );
     context.line(
       "Also, if your app has authentication, don't forget to set environment variables:"
     );
     context.line('* for local testing, edit the .env file');
-    context.line('* for produciton, use `zapier env` command');
+    if (!isVisual) {
```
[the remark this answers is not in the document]
xavdid: I think I agree. The biggest upside we get as written is that it's not a breaking change and can be released sooner (originally we promised that we would only do breaking changes every so often). 

So I think we have a few options here: 

1. leave it as is, which forces the dev to be explicit about which version they want
2. provide the `--latest` flag, which needs to be provided instead of the `--version` flag. Either denotes the command as referring to a Godzila app. In a breaking change, we could require a flag to explicitly type the app (`--visual` or `--legacy`)
3. Do a more full featured picker (similar to `zapier link`). 

Do you have strong feelings here? I feel like the UX as written is "good enough", especially since devs wont be running this often.

src/utils/convert.js
```diff
@@ -408,21 +445,30 @@ const writeGitIgnore = async newAppDir => {
...
+  // need to bump babel so this works
+  // const { silent, legacy } = { ...defaultOpts, ...opts };
```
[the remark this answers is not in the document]
xavdid: for sure. Our current version of babel throws an error here, so I think i'll:

* merge #430 
* merge develop back into this PR, 
* uncomment this

i'll do that after we're 👍 though to keep this PR cleaner. I think the babel one will make it loud

src/utils/convert.js
```diff
@@ -448,16 +494,45 @@ const convertApp = async (
...
+  promises.push(
+    legacy
+      ? writeLegacyZapierAppRc(newAppDir)
+      : writeVisualZapierAppRc(newAppDir, appInfo.id)
+  );
 
   return await Promise.all(promises);
 };
 
+const convertVisualApp = async (
```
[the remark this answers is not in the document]
xavdid: I'll do you one better - we can `_.partialRight` and not move any of the args!

xavdid:
ok, addressed most of those. let me know what you feel strongly about and we can get this over the line!

src/utils/convert.js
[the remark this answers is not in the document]
stevelikesmusic: And boom goes the dynamite 💥

src/commands/convert.js
stevelikesmusic: I hear you. I wonder if that's what most do?

It'd be nice to have the option to just run `zapier convert`, get a list of integrations, pick the integration I want, and confirm things like version and/or location. Not for this PR—just thinking out loud.

xavdid: totally. I think this is good to get it out and we can revisit it at some point.

src/utils/convert.js
```diff
@@ -162,19 +162,22 @@ const renderStep = (type, definition) => {
...
-        );
-      }
+      // Godzilla currently doesn't allow mutliple dynamic fields (see PDE-948) but when it does, this will account for it
+      let funcNum = 0;
+      fields.forEach((maybeFunc, index) => {
+        if (maybeFunc.source) {
+          const args = maybeFunc.args || ['z', 'bundle'];
+          // always increment the number, but only return a value if it's > 0
+          const funcName = `get${_.upperFirst(key)}${
+            funcNum ? funcNum++ : ++funcNum && ''
```
stevelikesmusic: Cool. So it would go as `getInputFields`, `getInputField1`, `getInputFields2`...

src/commands/convert.js
```diff
@@ -2,11 +2,7 @@ const utils = require('../utils');
 const constants = require('../constants');
 
 const convert = (context, appid, location) => {
-  context.line('Welcome to the Zapier Platform! :-D');
```
stevelikesmusic: It would be really cool here to clear out the location folder if it exists. I forgot to pass version to the tool, and it converted my vb app as if it was WB. When I ran it again, it just added to folder rather than replace it.

stevelikesmusic:
Do we still have a separate ticket to build out the interface for this command? I thought there was also a ticket that would output a list of integration you had before executing the convert.

xavdid:
We still have a separate ticket, but a lot of that work is in this one. 

We can launch this as is ASAP, but for the next major we could change the UI to have a nice selection setup. So many plates in the air!

src/utils/convert.js
xavdid: yep! read then increment. 

```
> let i = 0
undefined
> console.log(i++)
0
undefined
> i
1
> i
1
> console.log(i++)
1
undefined
> i
2
> console.log(++i)
3
undefined
> i
3
> console.log(++i)
4
undefined
> i
4
```

src/commands/convert.js
xavdid: good call out. there's another place we check for an empty folder- we can borrow that behavior.

xavdid:
Ah, so it turns out sorting keys is possible, but non-trivial if you have nested objects (which we do, in spades): https://stackoverflow.com/a/16168003/1825390

I'll skip for now but it'd be a nice cleanup thing to come back to (or render manually sometime). I'm going to go ahead and merge this as is.

PR #429, opened 2019-05-10
Add eslint-plugin-mocha

stevelikesmusic:
Adding `eslint-plugin-mocha` to enforce some linting rules in our tests.

One pain I've found is writing tests with exclusivity like `it.only` and accidentally commit it. We could also enforce some more restrictions in tests if we wanted like `it.skip` from making it into PRs. For now, exclusivity is the biggest pain.

Also, 🔤 our eslint rules.

To test it out, add `.only` to a test case. Your editor should light it up.


.eslintrc
```diff
@@ -9,11 +10,8 @@
     "Promise": true
   },
   "rules": {
-    "quotes": [2, "single", "avoid-escape"],
-    "strict": [0, "never"],
+    "mocha/no-exclusive-tests": "error",
```
stevelikesmusic: This is the only thing that was added. The rest is a resort

stevelikesmusic:
Thanks, @xavdid. Does app improvements have that standard documented anywhere? I'm curious what the overlaps and differences are between what they're doing and what FE is doing as whole. It'd be nice to adopt one standard for all of our JavaScript (or at least as much is possible)

I'm all for moving to jest as well. I think could gain by using `zapier-scripts` in the project at the least. One thing I'm looking to add for our Zapier's eslint plugin is the same change in this PR but for jest.

xavdid:
no problem! app improvements talks about their tools here: https://zapier.quip.com/BrO3ALZa5uYk/How-To-Make-a-CLI-App#ILNACAGivob

the FE team can do what they want, but it would be great for no one to need to maintain a detailed eslintconfig. but yeah, I may change some of that stuff when we move to a monorepo.

PR #430, opened 2019-05-10
try removing babel

xavdid:
We added babel when we wanted `async // await` and other nice features and then never really used them anyway (we still have a ton of promise code). 

Nowadays we're not using anything that's not in native node 8, so I figure'd we'd simplify our setup a little! 

Everything seems good with some cursory testing.


xavdid:
I think we can update npm separately from node, so we could stick to later versions. in any case, it'd like to be on yarn anyway, so it's probably a good move. that'll come with the monorepo swap, i think.

PR #432, opened 2019-05-16
Run app review checks before activation of apps

kola-er:
Run app review checks for non-public/non-pending apps after passing the validation checks on promote. This tells devs the outcome of the app review checks we run internally whenever they request for activation. This would reduce the back-and-forth of the app review process as there will be little to no request for change if users fixed all the issues before requesting for activation.

### DIFF
--- a/src/commands/promote.js
+++ b/src/commands/promote.js
@@ -2,6 +2,19 @@ const colors = require('colors/safe');
 const utils = require('../utils');
 const _ = require('lodash');
 
+const runAppReviewChecks = async (appId, appVersion, context) => {
+  context.line('\nRunning App Review Checks.\n');
+  const url = `/apps/${appId}/versions/${appVersion}/app-review-run`;
+  const response = await utils.callAPI(url, true);
+  const errorList = response.failed || [];
+  if (errorList.length > 0) {
+    const message = `Promotion failed for the following reasons:\n\n${errorList
+      .map(e => `* ${e.message}`)
+      .join('\n')}\n`;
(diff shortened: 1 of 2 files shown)


src/commands/promote.js
```diff
@@ -2,6 +2,19 @@ const colors = require('colors/safe');
 const utils = require('../utils');
 const _ = require('lodash');
 
+const runAppReviewChecks = async (appId, appVersion, context) => {
+  context.line('\nRunning App Review Checks.\n');
+  const url = `/apps/${appId}/versions/${appVersion}/app-review-run`;
+  const response = await utils.callAPI(url, true);
+  const errorList = response.failed || [];
+  if (errorList.length > 0) {
+    const message = `Promotion failed for the following reasons:\n\n${errorList
```
xavdid: I was going back and forth between whether this should say `Activation failed` or `Promotion failed`, since if they see it, they were trying to activate. Saying `promotion` make it sound like there was an error with promotion rather than it being blocked. 

Your call, not sure we even need a change here.

xavdid:
@kola-er what's the status of this PR? will it be merged soon?

kola-er:
Ahh...Sorry, I missed your last comment. Since the activation form in the web UI needs to be filled for CLI apps too when ready to go public, I thought it made more sense making the API call to run the app review checks before displaying the activation form on the activation form page.

So I'm closing this. Thanks for the review though.

kola-er:
@xavdid ^

PR #442, opened 2019-06-25
add archive redirect

xavdid:
If approved, i'll make identical changes on all other public repos and then archive them.

I've moved all the issues and the two PRs open here are mine and i'll make sure they're copied when appropriate


eliangcs (approved):
Looks good!

